PR #2, opened 2020-10-07
Initial import from jwst.stpipe

eslavich:
This is the initial import of code from jwst.  Similar to stdatamodels, I tried to make minimal changes and simply commented out code that referenced jwst-specific datamodels and steps.  I did not include anything from the scripts directory, so we'll need to follow up with that if we want the CLI tool to live in this package.

The first commit contains everything copied from jwst, so diffing off of that will show what needed to change.  Here's a compare view for that diff: https://github.com/eslavich/stpipe/compare/325a634...AL-406-initial-jwst-import-ii


jdavies-st (approved):
Looks good.

tests/steps/__init__.py
```diff
@@ -1,9 +1,10 @@
-from jwst.stpipe import (Pipeline, Step)
-from jwst import datamodels
-from jwst.datamodels import (
-    ImageModel,
-    ModelContainer,
-)
+from stpipe import (Pipeline, Step)
```
jdavies-st, line 1: I really hate how we have everything in `__init__.py` in this test dir.  Would be nice to refactor this in the future to be more standard.

tests/test_asdf_parameters.py
```diff
@@ -2,11 +2,11 @@
...
 
-from jwst.stpipe.config_parser import ValidationError
-from jwst.stpipe.step import Step
+from stpipe.config_parser import ValidationError
+from stpipe.step import Step
 
-from .steps import MakeListStep, MakeListPipeline
-from .util import t_path
+from steps import MakeListStep, MakeListPipeline
+from util import t_path
```
jdavies-st, line 9: Shouldn't these remain relative imports as these modules are in the `tests` dir?

tests/test_step.py
```diff
@@ -7,31 +7,34 @@
...
+# TODO: We don't have access to this here:
+# from jwst.refpix import RefPixStep
+from stpipe import Step, crds_client
+from stpipe import cmdline
+from stpipe.config_parser import ValidationError
 
-from .steps import EmptyPipeline, MakeListPipeline, MakeListStep, ProperPipeline
-from .util import t_path
+from steps import EmptyPipeline, MakeListPipeline, MakeListStep, ProperPipeline
+from util import t_path
```
jdavies-st: Same here.  Relative imports?

setup.cfg
```diff
@@ -38,3 +38,18 @@ test =
...
+exclude =
+    docs,
+ignore = E203, W503, W504, W605
+
+[tool:pytest]
+minversion = 4.6
+doctest_plus = true
+doctest_rst = true
+text_file_format = rst
+addopts = --show-capture=no --open-files
```
jdavies-st: You can remove `--show-capture=no` and add a newline.

Should probably remove this option in `jwst` as well.

tox.ini
```diff
@@ -0,0 +1,45 @@
+[tox]
+envlist= py38,style,bandit
+
+[testenv]
+extras= test
+# astropy will complain if the home directory is missing
+passenv= HOME
+commands=
+    pytest --open-files tests
```
jdavies-st, line 9: `--open-files` is already in `setup.cfg`, so probably not needed here?

setup.cfg
```diff
@@ -33,6 +33,7 @@ where = src
 
 [options.extras_require]
 test =
+    ci-watson>=0.3.0
```
jdavies-st, line 36: `ci-watson` should not be necessary for this package.  If it is, we should remove it.  This is only needed (really) for `jwst` regression testing access to Artifactory.  We have got sloppy here, and used one or two of the fixtures outside of the regression testing.  We should fix those.

src/stpipe/log.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, name, handler=None, level=logging.NOTSET,
         self.name = name
         self.handler = handler
         if not isinstance(self.handler, list):
-            if self.handler.strip() == '':
+            if self.handler is None or self.handler.strip() == '':
```
jdavies-st, line 111: What does this fix?  I wonder if this needs to be backported to `jwst.stpipe`?

setup.cfg
```diff
@@ -19,9 +19,22 @@ zip_safe = True
 python_requires = >=3.6
 setup_requires =
     setuptools_scm
+install_requires =
+    asdf>=2.7.1
+    crds>=7.4.1.3
+    astropy>=4.0
+    stdatamodels @ git+https://github.com/spacetelescope/stdatamodels
```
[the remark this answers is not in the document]
eslavich: We've claimed the package name, but I figured during this initial period of rapid development we might as well install from master to save from having to make a bunch of 0.x releases.

tests/test_asdf_parameters.py
eslavich: The tests directory isn't a Python package, so relative imports don't work here.  Python isn't able to resolve the `.` into anything.

tests/steps/__init__.py
eslavich: I agree, this needs cleanup.

setup.cfg
eslavich: This provides the `_jail` fixture, which is used in a handful of tests.

tox.ini
eslavich: Yes good point!

tests/test_asdf_parameters.py
jdavies-st: Ah ok!

setup.cfg
jdavies-st: Hmm.

We should really just be using `tmp_path` in those tests I think.  `_jail` is just a lazy way to get it, sorta.

src/stpipe/crds_client.py
```diff
@@ -42,8 +42,11 @@
 from crds.core import config, exceptions, heavy_client, log
 from crds.core import crds_cache_locking
 
-from ..lib import s3_utils
-from ..datamodels import open as dm_open
+from . import s3_utils
```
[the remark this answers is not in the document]
eslavich: It was shared between stdatamodels and stpipe.  But you raise a valid question: why not just import it from stdatamodels?  Answer: cuz I didn't think of that.

src/stpipe/step.py
```diff
@@ -29,14 +29,24 @@
...
+# from ..associations.lib.update_path import update_key_value
+
+# TODO: Neither ModelContainer nor StepParsModel are available yet:
+from stdatamodels import DataModel
+# from ..datamodels import (DataModel, ModelContainer, StepParsModel)
+
+# TODO: The open function is not yet available:
+# from ..datamodels import open as dm_open
+
+from .class_property import ClassInstanceMethod
```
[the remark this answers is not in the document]
eslavich: I suppose if we wanted to we could create another package of general purpose Python utilities and share that around.  It doesn't seem to "belong" to any of these packages, exactly.

setup.cfg
eslavich: I'll add an issue to clean this up

src/stpipe/log.py
eslavich: The `handler` attribute is set to None in the tox environment.  I didn't take the time to figure out why, but since jwst doesn't run tests in tox it's not had this issue.

src/stpipe/crds_client.py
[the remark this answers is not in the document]
eslavich: (this has been updated to fetch s3_utils from stdatamodels)

PR #11, opened 2021-02-04
Update to current state of jwst stpipe-transition branch

eslavich:
This brings us current with jwst commit f7dbb5c31906ef4826319eb0f2425380a8310b96.  The package has very poor test coverage at this point, so I'm checking it using the jwst regression tests here:

https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/60

(The one failure is due to an import I missed in the jwst code which I've since fixed).

[This](https://github.com/eslavich/jwst/compare/f7dbb5c...eslavich:eslavich-test-with-stpipe) shows my jwst dev branch with stpipe code removed.


jdavies-st (approved):
This looks good.

In a future PR, we may want to do some downstream tests of `jwst` and `romancal` unit tests as we do for `stdatamodels` and `asdf`.

PR #12, opened 2021-02-08
Create StpipeExitException for exiting CLI with a specific exit status

eslavich:
@jdavies-st pointed out that `NoDataOnDetectorError` is JWST-specific and probably not useful to keep in stpipe.  ~~This removes that exception from this package and adds more general handling of any `SystemExit` subclass.~~  This removes that exception and adds an `StpipeExitException` base class that jwst can use to cause the CLI to exit with a particular status code.

### DIFF
--- a/scripts/strun
+++ b/scripts/strun
@@ -5,15 +5,15 @@ Run Steps from the command line
 
 Exit Status
 -----------
-    0:  Step completed satisfactorally
+    0:  Step completed satisfactorily
     1:  General error occurred
-    64: No science exists
+
+Other status codes are Step implementation-specific.
 """
 
 import sys
 
 import stpipe
-from stpipe.exceptions import NoDataOnDetectorError
 from stpipe import Step
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


scripts/strun
```diff
@@ -24,20 +24,12 @@ if __name__ == '__main__':
...
-    except NoDataOnDetectorError:
-        #  No science data is present on a detector. This can happen
-        #  with NIRSpec and the NRS2 detector: Situations occur when no spectra
-        #  disperse across both detectors.
-        #
-        #  Special exit status is provided so that automatic processing can detect
-        #  this situation and act accordingly.
-        #
-        #  Full discussion: https://github.com/spacetelescope/jwst/issues/2336
+    except SystemExit as e:
```
jdavies-st, line 27: `e` is not being used in this `except` block.

And since `SystemExit` subclasses `BaseException` instead of `Exception`, it's possible we don't even need this `except` block.  I.e. the `except Exception` block below won't catch `SystemExit`, so it will be raised from the step code regardless.

eslavich: Oh hey so it does!  Thanks, I'll fix that up.

eslavich: (now updated to define and catch an `StpipeExitException`)

jdavies-st: I guess the one advantage of capturing it here and printing the traceback is that normally when `SystemExit` is raised, it just exits, without printing the traceback.  At least that's the description here:

https://docs.python.org/3/library/exceptions.html#SystemExit

🤔

eslavich: I played around with this and noticed that catching `SystemExit` results in tracebacks when argparse prints its help output, since it raises `SystemExit` to bail out in that case.

jdavies-st (approved):
Looks really good!  I think we're ready to merge and release!

PR #14, opened 2021-02-08
Add infrastructure for new CLI + implementation of 'stpipe list' command

eslavich:
This adds a little framework for CLI commands and implements `stpipe list`, which lists step and pipeline classes registered with the entry point.

### DIFF
--- a/src/stpipe/__main__.py
+++ b/src/stpipe/__main__.py
@@ -2,5 +2,5 @@ from .cli.main import main
 
 
 # main() always raises SystemExit, so we don't need
-# do perform any exception handling here.
+# to perform any exception handling here.
 main()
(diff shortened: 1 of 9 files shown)


stscieisenhamer (approved):
Presuming that the appropriate exit statuses will be produced, LGTM.

src/stpipe/__main__.py
```diff
@@ -0,0 +1,6 @@
+from .cli.main import main
+
+
+# main() always raises SystemExit, so we don't need
+# do perform any exception handling here.
```
jdavies-st, line 5:
```suggestion
# to perform any exception handling here.
```

jdavies-st (approved):
This is really good.  I like how you're keeping the CLI separate from the library code.

Hopefully in the future we can then pull the logging handlers over to the CLI side as well, and remove them from the library code.  😁

PR #17, opened 2021-02-16
change default value of output_ext

nden:
This change allows subclasses to define their own `output_ext`. The extension is used to define what output data format to use.

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -43,7 +43,7 @@ class Step:
     post_hooks         = string_list(default=list())
     output_file        = output_file(default=None)   # File to save output to.
     output_dir         = string(default=None)        # Directory path for output files
-    output_ext         = string(default='.fits')     # Default type of output
+    output_ext         = string()                    # Default type of output
     output_use_model   = boolean(default=False)      # When saving use `DataModel.meta.filename`
     output_use_index   = boolean(default=True)       # Append index.
     save_results       = boolean(default=False)      # Force save results
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+0.2.0 (Unreleased)
+==================
+
+- Remove the default value of ``output_ext`` so subclsses can define it. [#17]
+  
 0.1.0 (2021-02-08)
 ==================
 


src/stpipe/step.py
```diff
@@ -43,7 +43,7 @@ class Step:
     post_hooks         = string_list(default=list())
     output_file        = output_file(default=None)   # File to save output to.
     output_dir         = string(default=None)        # Directory path for output files
-    output_ext         = string(default='.fits')     # Default type of output
+    output_ext         = string(default=None)     # Default type of output
```
eslavich, line 46: Can we remove the default parameter entirely?  This will result in the following error if someone subclasses `Step` but doesn't specify `output_ext`:

```python
In [4]: from stpipe.step import Step

In [5]: class StepSubclass(Step):
   ...:     pass
   ...:

In [6]: StepSubclass()

ValidationError: Config parameter 'output_ext': missing
```
so it forces the user to specify `output_ext`:
```python
In [7]: StepSubclass(output_ext='.asdf')
2021-02-16 09:40:09,825 - stpipe.StepSubclass - INFO - StepSubclass instance created.
Out[7]: <__main__.StepSubclass at 0x7fb505340d30>
```
Alternatively the subclass author can override the `output_ext` definition with a default:
```python
In [8]: class StepSubclassWithDefault(Step):
   ...:     spec = """
   ...:     output_ext         = string(default='.asdf')     # Default type of output
   ...:     """

In [9]: StepSubclassWithDefault()
2021-02-16 09:41:34,389 - stpipe.StepSubclassWithDefault - INFO - StepSubclassWithDefault instance created.
Out[9]: <__main__.StepSubclassWithDefault at 0x7fb5053d1040>
```

eslavich: By "remove the default parameter" I mean just remove the `default` argument to the `string()` function:
```python
output_ext         = string()                    # Default type of output
```

nden: This was causing an error in jwst but let me check again, it's possible my env was not correct.

nden: Sorry, it was failing in romancal with the error above. I was hesitant to add `configobj` to romancal.

eslavich (approved):
LGTM

PR #20, opened 2021-03-22
added changes to support other data model implementation for romancal

### LINKED ISSUE #19: Support other datamodels than provided by stdatamodels
Currently the plan is to have romancal use tagged ASDF files, but doing so requires a different class implementation for data models. Stpipe should be able to accommodate different datamodel implementations without requiring subclassing of the stdatamodel implementation. Currently the only dependence on the datamodel class by stpipe is on these methods: 

- DataModel.crds_observatory
- DataModel.get_crds_parameters
- DataModel.save
It also uses the attribute meta.filename

perrygreenfield:
stpipe.py now defines DataModel as an abstract class that tests a prospective subclass for the methods that stpipe depends on.

This closes #19


perrygreenfield:
Further changes: moved ABC DataModel class to its own file so that it can be used by both the config_parser and setp.

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    @classmethod
+    def __subclasshook__(cls, C):
+        """
+        Psuedo subclass check based on these attributes and methods
+        """
+        if cls is DataModel:
+            mro = C.__mro__
+            if (any([hasattr(CC, "crds_observatory") for CC in mro]) and
```
eslavich, line 18: What do we want to do about `DataModel.meta.filename`?  AFAICT, that's the only other model attribute accessed by stpipe.  It may be possible to remove references to it but at a glance that doesn't seem trivial.  Maybe we need a new ABC method or property that retrieves the filename?  It doesn't look like stpipe needs to modify it.

perrygreenfield: I wondered about this, but it wasn't obvious to me how to deal with it since `meta` isn't obviously a requirement of `DataModel`, but imposed by a schema. In the end I figured that the requirement would be implicit, but if you have a good way to deal with it, I'm happy to use it. So add a property to all `DataModel`s?

src/stpipe/step.py
```diff
@@ -24,16 +25,48 @@
...
 from . import config
 from . import config_parser
 from . import crds_client
 from . import log
 from . import utilities
+from .datamodel import DataModel
 from .format_template import FormatTemplate
 
 
+# class DataModel(abc.ABC):
```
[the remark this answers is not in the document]
perrygreenfield: True.

src/stpipe/datamodel.py
eslavich: I think that would work out nicely, something like:

```
@abc.abstractmethod
def get_filename(self):
  pass
```

Then in JwstDataModel / RomanDataModel that can be implemented like:

```
def get_filename(self):
  return self.meta.filename
```

(I'm not sure about property vs method, but the get_ is nice because it has less of a chance of colliding with a schema property).

perrygreenfield: This would be good, but for the sake of painless transitions, I'd suggest making the changes first for JWST and romancal, then here when they support this method.

src/stpipe/step.py
```diff
@@ -24,16 +24,48 @@
...
 except ImportError:
     DISCOURAGED_TYPES = None
-from stdatamodels import DataModel
 
 from . import config
 from . import config_parser
 from . import crds_client
 from . import log
 from . import utilities
+from .datamodel import DataModel
```
jdavies-st, line 33: This import can be removed.

src/stpipe/step.py
```diff
@@ -24,16 +24,48 @@
...
 from .format_template import FormatTemplate
 
 
+# class DataModel(abc.ABC):
+#     """
+#     This Abstract Base Class is intended to cover multiple implmentations of
+#     data models so that each will be considered an appropriate subclass of this
```
jdavies-st, line 40: This commented-out code can be removed as well.

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,34 @@
...
+                any([hasattr(C, "save") for CC in mro])):
+                return True
+        return False
+
+    @abc.abstractmethod
+    def crds_observatory(self):
+        pass
+
+    @abc.abstractmethod
+    def get_crds_parameters(self):
```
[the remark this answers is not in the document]
perrygreenfield: I've attempted to add that.

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,34 @@
...
+        pass
+
+    @abc.abstractmethod
+    def save(self, path, dir_path=None, *args, **kwargs):
+        pass
```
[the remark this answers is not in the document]
perrygreenfield: apparently it didn't or was rendered irrelevant by a subsequent change.

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,34 @@
+
+import abc
+
+class DataModel(abc.ABC):
```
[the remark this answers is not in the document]
perrygreenfield: BruceAlmighty? But really, DataModelABC?

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def __subclasshook__(cls, C):
+        """
+        Psuedo subclass check based on these attributes and methods
+        """
+        if cls is DataModel:
+            mro = C.__mro__
+            if (any([hasattr(CC, "crds_observatory") for CC in mro]) and
+                any([hasattr(C, "get_crds_parameters") for CC in mro]) and
```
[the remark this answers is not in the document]
perrygreenfield: Yeah, sloppy on my part.

src/stpipe/config_parser.py
```diff
@@ -9,7 +9,7 @@
 
 from asdf import open as asdf_open
 from asdf import ValidationError as AsdfValidationError
-from stdatamodels import DataModel
+from .datamodel import DataModel
```
nden, line 12: Move this down to the internal imports.

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,58 @@
+
+import abc
+
+class DataModel(abc.ABC):
+    """
+    This Abstract Base Class is intended to cover multiple implmentations of
```
nden, line 6:
```suggestion
    This Abstract Base Class is intended to cover multiple implementations of
```

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+
+    @abc.abstractmethod
+    def save(self, path, dir_path=None, *args, **kwargs):
+        """
+        Save to a file.
+
+        Parameters
+        ----------
+        path : string or func
```
nden, line 44:
```suggestion
        path : str or callable
```

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,58 @@
...
+            File path to save to.
+            If function, it takes one argument with is
```
nden, line 46:
```suggestion
            If function, it takes one argument which is
```

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,58 @@
...
+            model.meta.filename and returns the full path string.
+
+        dir_path: string
```
nden, line 49:
```suggestion
        dir_path : str
```

src/stpipe/datamodel.py
```diff
@@ -0,0 +1,34 @@
...
+        if cls is DataModel:
+            mro = C.__mro__
+            if (any([hasattr(CC, "crds_observatory") for CC in mro]) and
+                any([hasattr(C, "get_crds_parameters") for CC in mro]) and
+                any([hasattr(C, "save") for CC in mro])):
+                return True
+        return False
+
+    @abc.abstractmethod
+    def crds_observatory(self):
```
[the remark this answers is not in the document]
nden: Just for my understanding, isn't `@abstractproperty` deprecated, the recommendation being to use `@abstractmethod` with `property`?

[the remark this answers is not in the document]
eslavich: That's news to me!  Since Python 3.3, apparently.  Looks like I have some decorators to update...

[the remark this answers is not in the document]
perrygreenfield: Just to clarify, what changes if any should I make?

src/stpipe/datamodel.py
jdavies-st: Even better than a method would be a filename property.  I suspect we will never want to pass an argument to get the filename.

eslavich: Given that we're treating all schema properties as `DataModel` instance attributes, it seems safest to namespace the ABC methods somehow to help prevent collisions.  The method names starting with `get_` are a natural way to do that...

eslavich: (we may want to change `crds_observatory` to a method for the same reason)

nden:
What is needed to finalize this?

perrygreenfield:
That’s a little unclear now. I have just finished updating the tests and moving some of them out of romancal to rad and roman_datamodels. The complicating factor is all these changes going into rad schemas that presumably will need to be merged with my version of rad, and any changes need to support those changes. 

I figured I should finish the test changes, which I have just done, and then figure out how to merge all this stuff. It could get somewhat messy. I’m not sure what work is involve in that. As far as tickets go, I think we can close this jira, ticket and create a new one to incorporate the changes requested by azure. Does that make sense? I may want to wait and see  what changes are being made before continuing. Doing things at the same time seems to be asking for trouble.

Perry

src/stpipe/datamodel.py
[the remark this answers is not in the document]
eslavich: This should be:

```python
@property
@abc.abstractmethod
def crds_observatory(self):
```

src/stpipe/datamodel.py
eslavich: Added an issue to address this later: https://github.com/spacetelescope/stpipe/issues/21

jdavies-st (approved):
This looks good to me.  I ran the `jwst` regression tests using this branch, and all is looking good.

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/240/

nden (approved):

PR #22, opened 2021-04-26
JP-1940: Change method to combine kwargs with provided cfg args in .call()

tapastro:
Fixes [jwst #5789](https://github.com/spacetelescope/jwst/issues/5789)
Resolves [JP-1940](https://jira.stsci.edu/browse/JP-1940)

Not sure if this is the correct procedure, but a JP ticket filed with a jwst repo issue has been traced to an issue in stpipe - specifically, if a user provides both a config_file and specific step flags, the merging of the step flag dicts was not a deep merge but an overwrite. This is easily fixed by using a different, existing method that ensures nested dicts are appropriately merged.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+0.3.0 (unreleased)
+==================
+
+- Change ConfigObj.update() to merge() when combining user-provided
+  config_file and step-specific flags during a step.call() to properly
+  merge dicts of step flags [#22]
+
 0.2.0 (2021-04-22)
 ==================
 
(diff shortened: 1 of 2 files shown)


jdavies-st (approved):
This looks good.

Nice to know that `update()` for dicts is not recursive.  I didn't know that!

Needs a changelog.

tapastro:
> This looks good.
> 
> Nice to know that `update()` for dicts is not recursive. I didn't know that!
> 
> Needs a changelog.

New to this domain, should I be putting the new entry under a new header or under 0.2.0?

jdavies-st:
@eslavich we should probably start doing milestones for major `jwst` or `romcal` releases in these shared libraries.  That might help to know when we need releases.  That said, for `jwst`, this should probably be released in the next couple weeks so it can make it into B7.8

PR #26, opened 2021-06-28
Prevent ConfigObj from treating DataModel as a config section

eslavich:
Over in stdatamodels I'm working on making DataModel implement collections.abc.Mapping, but that causes ConfigObj to treat reference overrides as config sections instead of config values.  This PR updates the merge_config function so that only true dicts are treated as config sections.


jdavies-st (approved):
🚀

PR #29, opened 2021-10-04
JP-1793: Expose formatting of output file to Step instances

tapastro:
Addresses JP-1793

This PR adds a new boolean to the Step spec, allowing for a Step instance to set save_model(format=False), formatting the output_model.meta.filename accordingly.

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -46,6 +46,7 @@ class Step:
     output_ext         = string()                    # Default type of output
     output_use_model   = boolean(default=False)      # When saving use `DataModel.meta.filename`
     output_use_index   = boolean(default=True)       # Append index.
+    output_use_format  = boolean(default=True)       # Apply default formatting rules to filename
     save_results       = boolean(default=False)      # Force save results
     skip               = boolean(default=False)      # Skip this step
     suffix             = string(default=None)        # Default suffix for output files
@@ -448,10 +449,16 @@ class Step:
                     if len(results_to_save) <= 1:
                         idx = None
                     if isinstance(result, AbstractDataModel):
-                        self.save_model(result, idx=idx)
+                        if self.output_use_format:
+                            self.save_model(result, idx=idx)
+                        else:
+                            self.save_model(result, idx=idx, format=False)
                     elif hasattr(result, 'save'):
                         try:
-                            output_path = self.make_output_path(idx=idx)
+                            if self.output_use_format:
+                                output_path = self.make_output_path(idx=idx)
+                            else:
+                                output_path = self.make_output_path(idx=idx, format=False)
                         except AttributeError:
                             self.log.warning(
                                 '`save_results` has been requested,'


jdavies-st:
Is this parameter something that needs to be exposed to the user, i.e. the person who runs a pipeline, or to the programmer, i.e. the person writing pipeline steps?  If only the latter, it would be best to have it just be defined internally as a class instance attribute, but not something that is part of the public API via `spec`.  Thoughts @stscieisenhamer ?

tapastro:
It might be convenient to a pipeline user if they'd prefer to set their own naming scheme/format in the association product 'name' field, which I assume would be paired with the use of output_use_model; that's the only use case that comes to my mind right now.

jdavies-st:
Is there a reason to have `format=False` at all?  Or at least as the default?  If the output product name has a format string `some_name_{suffix}-foo_bar` shouldn't it always format?

tapastro:
I'm not sure I have the 'sign' of my booleans in order, but I think we've flipped here - before adding this parameter and using the default format, the output file retains the {suffix} string without substitution. Setting format=False uses the built-in formatting to place the suffix inside the curly braces. This seems to follow the groundwork you guys laid out almost a year ago [here](https://github.com/spacetelescope/jwst/pull/5519/files#r537785329), unless I've misunderstood.

jdavies-st:
I guess what I'm asking is, if there's no `{suffix}` in the file name, can't we just do normal append `_suffix` to the filename?  If there is `{suffix}` in the association product name, then we do the other thing?  And there's no need to have a `format` parameter to `self.save_model` or this new top-level attribute as well?

Mo parameters, mo confusion.  Can we get rid of parameters instead of adding more?

tapastro:
> I guess what I'm asking is, if there's no `{suffix}` in the file name, can't we just do normal append `_suffix` to the filename? If there is `{suffix}` in the association product name, then we do the other thing? And there's no need to have a `format` parameter to `self.save_model` or this new top-level attribute as well?
> 
> Mo parameters, mo confusion. Can we get rid of parameters instead of adding more?

I don't disagree with any of this, though this seems like 'secret' behavior that we would need to document well for it to be used.

tapastro:
I also still have some confusion on what format = True vs. None vs. False results in, though I'm trying to parse stpipe.step to figure this out. It appears that None utilizes the default format, False uses the basename, but it looks like True does something else?

stscieisenhamer:
Background: The ultimate parameter in question that all this leads to is [_make_output_path name_format](https://github.com/spacetelescope/stpipe/blob/caf3775014989574e27eda7cdb85909a9ab553c7/src/stpipe/step.py#L979). At the time, due to much bigger issues (Step frankly didn't work), the stopgap solution was to make `name_format` a tri-state parameter.  Note that the ultimate goal of `_make_output_path` is to get a suffix on the output file name and ensure, to the best of its ability, to make the name unique amongst other output names. As such, the comment for `name_format` should really be:

String defining the format of the output name which defines how `**components` are inserted into the output file name. If `None`, use the default formatting, which is to simply append `Step.suffix` to the name.  However, if `False`, use `basepath` as its own format string.

Ideally, this should be refactored to remove the tri-state (None, False, string). To do so will require a bit of introspection and will not necessarily solve the immediate bug.

The immediate bug is that the Step infrastructure does not have any way of modifying `name_format`, regardless of the tri-state issue. Hence my initial recommendation to make this a global `spec`. I do agree that this unnecessarily crowds up the spec list. Making this a class attribute would be sufficient.  The default would be `Step.name_format = None`. Then, if a specific instance wanted to do what WfsCombineStep is doing, it can set itself to `False`. This would remove the `if...else` in lines 452 and 458 in `run` to just make the statement be `output_path = self.make_output_path(idx=idx, name_format=self.name_format)`

Another fix would be to fix the argument naming between `Step.save_model` where this is `format` as opposed to the deeper call where for `make_output_path` it is `name_format`. Again, at the time, bigger fish to fry. This confusion makes errors like I see in the current commit in the `run` method, line 461, where it is mistakenly using `format` instead of `name_format` for the `make_output_path` call. Again, a possible future refactor.

stscieisenhamer (approved):

src/stpipe/step.py
```diff
@@ -56,6 +56,10 @@ class Step:
     # name.  Must be globally unique!
     class_alias = None
 
+    # Default formatting of output filenames unless overridden, typically with
+    # a format string provided along with output_use_model
+    name_format = None
```
jdavies-st, line 61: Maybe this comment can be closer to Jonathan's description:

"String defining the format of the output name which defines how **components are inserted into the output file name. If None, use the default formatting, which is to simply append Step.suffix to the name. However, if False, use basepath as its own format string."

jdavies-st (approved):
Looks good!

How `make_output_path` actually works is pretty opaque.  It would be nice to capture @stscieisenhamer 's comments in https://github.com/spacetelescope/stpipe/pull/29#issuecomment-935327823 in the actual documentation or docstrings, hence my comment below.

Agree this all needs a refactor, but this does fix the current issue, so 👍

jdavies-st:
For some reason required checks does not respect `ci_skip`.  No worries.  Merging.

PR #32, opened 2021-10-13
JP-579: Add parsing for logcfg in Step.call

tapastro:
Closes [#3217](https://github.com/spacetelescope/jwst/issues/3217)
Resolves [JP-579](https://jira.stsci.edu/browse/JP-579)

This PR adds a check for 'logcfg' provided in the kwargs of a Step.call() - if present, load the config file. The kwarg needs to then be removed from the config or it will fail validation.

EDIT: I'm not sure if this is the best place in the code to set the log.load_configuration and 'pluck' it out of the config/kwargs - perhaps I could do it in Step.from_config_section instead? Happy to hear suggestions.

### DIFF
--- a/tests/test_step.py
+++ b/tests/test_step.py
@@ -1,7 +1,6 @@
 """Test step.Step"""
 import pytest
-
-import numpy as np
+import logging
 import asdf
 
 import stpipe.config_parser as cp


src/stpipe/step.py
```diff
@@ -576,6 +576,15 @@ def call(cls, *args, **kwargs):
...
 
+        if 'logcfg' in config:
+            try:
+                log.load_configuration(config['logcfg'])
+            except Exception as e:
+                raise ValueError(
+                    f"Error parsing logging config {config['logcfg']}:\n{e}"
+                )
+            del config['logcfg']
+
```
jdavies-st, line 587: The traceback produced here will be a bit confusing.

Slightly better exception handling would be to raise the 2nd except `from` the first.  Then it will be clear what caused what in the traceback.  And you don't have to try to include the handled exception inside the one your'e raising.

Also, not sure `ValueError` is appropriate.  Maybe `RuntimeError` would be better?

```suggestion
            except Exception as e:
                raise RuntimeError(
                    f"Error parsing logging config {config['logcfg']}"
                ) from e
            del config['logcfg']

```

tests/test_step.py
```diff
@@ -177,3 +200,29 @@ def test_build_config_step_kwarg(mock_step_crds, config_file_step):
...
+    for log in logdict:
+        if not isinstance(logdict[log], logging.PlaceHolder):
+            for handler in logdict[log].handlers:
+                if isinstance(handler, logging.FileHandler):
+                    logdict[log].removeHandler(handler)
+                    handler.close()
+
+    if os.path.isfile(tmpdir / 'myrun.log'):
+        with open(tmpdir / 'myrun.log', 'r') as f:
+            fulltext = '\n'.join([line for line in f])
```
jdavies-st, line 226: Since this a test, we don't want `if` statements.  We want this test to fail if it can't find the log.  Otherwise it will fail silently here, and `fulltext` is not defined outside this `if` block.

```suggestion
    with open(tmpdir / 'myrun.log', 'r') as f:
        fulltext = '\n'.join([line for line in f])
```

CHANGES.rst
```diff
@@ -18,6 +18,8 @@
 
 - Fix wiping out substep parameters settings when using Step.call [#28]
 
+- Add keyword 'logcfg' to Step.call() to set logging configuration [#32]
```
nden, line 21: Stpipe 3.0 was released already. This needs to be moved to a new section.

tapastro: Sorry about that. Should I add a new section header? Will this cause merge rebasing for future stpipe PRs?

PR #33, opened 2021-10-15
Support list arguments

mcara:
This PR fixes a bug due to which list arguments are not supported. The bug is basically due to converting _everything_ to a string (`FromCommandLine`).

I am not sure which would be a preferred fix:

```python
if key in ['input_file', 'output_file']:
```
or
```python
if isinstance(val, str):
```

CC: @nden

### DIFF
--- a/tests/test_step.py
+++ b/tests/test_step.py
@@ -220,7 +220,9 @@ def test_build_config_step_kwarg(mock_step_crds, config_file_step):
 
 
 def test_step_list_args(mock_step_crds, config_file_list_arg_step):
-    """Test that local config overrides defaults and CRDS-supplied file"""
+    """ Test that list arguments, provided as comma-separated values are parsed
+        correctly.
+    """
     config, returned_config_file = ListArgStep.build_config(
         'science.fits',
         config_file=config_file_list_arg_step


src/stpipe/cmdline.py
```diff
@@ -120,7 +120,8 @@ def set_value(subconf, key, val):
             set_value(subconf.setdefault(root, {}), rest, val)
         else:
             val, comment = config._handle_value(val)
-            subconf[root] = FromCommandLine(val)
+            if val in ['input_file', 'output_file']:  # or isinstance(val, str)
```
stscieisenhamer, line 123: Not that I understand why `val` gets values such as 'input_file', but the `isinstance` just seems more general.

mcara: ah, not `val`. Should have been `key`.

mcara: My initial implementation was using `isinstance` but I thought that maybe we do not need to convert every parameter to `FromCommandLine` unless we know it is a file path.

If `isinstance` is preferable, I'll modify the code.

src/stpipe/cmdline.py
```diff
@@ -120,7 +120,8 @@ def set_value(subconf, key, val):
             set_value(subconf.setdefault(root, {}), rest, val)
         else:
             val, comment = config._handle_value(val)
-            subconf[root] = FromCommandLine(val)
+            if key in ['input_file', 'output_file']:  # or isinstance(val, str)
```
stscieisenhamer, line 123: Since the whole point of this seems to be to determine a working directory, and the presumption, in other parts of the Step I/O system, is that the working directory is determined by the input, methinks this logic is good, but only keying off of 'input_file', not both 'input_file' and 'output_file'. 

Thus the only parameter that will get this bizarre recast to `FromCommandLine` would be 'input_file'. That is a good thing methinks.

stscieisenhamer (approved):

src/stpipe/cmdline.py
mcara: `if key in ['input_file', 'output_file']` was too restrictive and resulted in numerous unit test failures in `jwst` due to incorrect path. Using `isinstance`, on the other hand, did not result in failures.

nden:
Can you come up with a test?

PR #35, opened 2021-11-10
Add Step.log_records property

eslavich:
This adds a the Step.log_records property which contains a list of logging.LogRecord emitted during the most recent run of the step.

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -295,6 +295,9 @@ class Step:
             variables on the new Step instance.
         """
         self._reference_files_used = []
+        # A list of logging.LogRecord emitted to the stpipe root logger
+        # during the most recent call to Step.run.
+        self._log_records = []
         self._input_filename = None
         self._input_dir = None
         self._keywords = kws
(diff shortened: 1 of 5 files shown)


src/stpipe/step.py
```diff
@@ -356,6 +359,17 @@ def _check_args(self, args, discouraged_types, msg):
                     "{0} {1} object.  Use an instance of AbstractDataModel instead.".format(
                         msg, i))
 
+    @property
+    def log_records(self):
```
kmacdonald-stsci, line 363: Is this a good name for this method, like `get_records` or `retrieve_records`?  This method name sounds like it's logging records, but the function returns records.

eslavich: Since it's a property it should be a noun, I think, but I'm open to other suggestions.  I didn't notice the similarity between "log_records" and "record_logs" until now...

kmacdonald-stsci (approved):

tapastro (approved):
LGTM

PR #37, opened 2022-03-01
JP-2498 For classmethods, use the delegator logger

stscieisenhamer:
Resolves issue [JP-2498](https://jira.stsci.edu/browse/JP-2498)

For `Pipeline` classmethods, use the `log.delegator` logger for messaging.


eslavich (approved):
LGTM

PR #50, opened 2022-03-31
JP-2520: Bugfix for ModelContainers

tapastro:
The modification made in #38 does not propagate metadata from ModelContainers, as the container itself contains no metadata with which to fetch references from CRDS. Taking from _precache_references() code below the code I added, I'm following in someone's footsteps by selecting ModelContainers on isinstance(input, Sequence). With this, I can grab the metadata from the single asn member (due to asn_n_members=1 in datamodel_open).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,9 @@
 
 - Pass header-only model to steps for CRDS fetching to reduce memory usage [#38]
 
+- Ensure product header is passed for CRDS fetching instead of empty
+  ModelContainer header [#50]
+
 0.3.1 (2021-11-12)
 ==================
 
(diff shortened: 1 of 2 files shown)


hbushouse (commented):
Looks reasonable to me, but I know almost nothing about stpipe, so ...

Needs a change log entry.

stscieisenhamer (approved):
As mentioned, since this solution is used in other places, LGTM.

A future thought: stdatamodels should have an abstract class/empty class, something like `ModelList`, that `ModelContainer` would subclass from that can then be used to flag such situations. Would not do this for this PR, but file a `stdatamodels` issue.

PR #56, opened 2022-06-02
move build configuration into `pyproject.toml`

zacharyburnett:
`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.


docs/source/conf.py
```diff
@@ -19,17 +17,17 @@ def setup(app):
...
+# Read the package's `pyproject.toml` so that we can use relevant
 # values here:
-conf = ConfigParser()
-conf.read(REPO_ROOT/"setup.cfg")
-setup_metadata = dict(conf.items("metadata"))
+with open(REPO_ROOT / "pyproject.toml", "rb") as configuration_file:
+    conf = tomli.load(configuration_file)
+setup_cfg = conf['project']
 
 project = setup_metadata["name"]
```
WilliamJamieson, line 28:
```suggestion
project = setup_cfg["name"]
```

This should fix the failing Docs job. A bigger question is should we just create a RTD and have RTD build the docs?

WilliamJamieson (changes requested):

zacharyburnett:
@WilliamJamieson I didn't realize that `stpipe` doesn't have a RTD site, I can make a PR to add that so we can get rid of the docs job in github actions

pyproject.toml
```diff
@@ -1,6 +1,73 @@
+[project]
+name = 'stpipe'
+description = 'Framework for calibration pipeline software'
+readme = 'README.md'
+requires-python = '>=3.7'
+license = { file = 'LICENSE' }
+authors = [{ name = 'STScI' }]
```
WilliamJamieson, line 7:
```suggestion
authors = [{ name = 'STScI', email = 'help@stsci.edu' }]
```

There is no email, which is breaking the docs build

WilliamJamieson (changes requested):

WilliamJamieson (approved):

PR #57, opened 2022-06-06
JP-2609: Fix ignored step-pars during cmdline pipeline run

tapastro:
Currently, running a pipeline while providing a pars-{step} file overwrites the user-supplied pars with the default pars. It appears as though a simple swap in a merge may be all that's required to get this functioning - the overwrite priority is now flipped for parameters which are specified in both the default spec and the user pars file.

Addresses [JP-2609](https://jira.stsci.edu/browse/JP-2609)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,10 @@
 
 - Update CI workflows to cache test environments and depend upon style and security checks [#55]
 
+- Correctly handle config merges of default spec, any possible step-pars files (from
+  CRDS or the user), and either command line (for strun) or step parameter dictionary (for interactive
+  session Pipeline.call()) parameter specifications [#57]
+
 0.3.3 (2022-04-07)
 ==================
 
--- a/src/stpipe/cmdline.py
+++ b/src/stpipe/cmdline.py
@@ -286,7 +286,7 @@ def just_the_step_from_cmdline(args, cls=None):
     # This is where the step is instantiated
     try:
         step = step_class.from_config_section(
-            config, name=name, cmd_args=args)
+            config, name=name, param_args=args)
     except config_parser.ValidationError as e:
         # If the configobj validator failed, print usage information.
         _print_important_message("ERROR PARSING CONFIGURATION:", str(e))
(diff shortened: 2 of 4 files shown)


tapastro:
Note: the simple trick failed.

WIP

tapastro:
@stscieisenhamer I committed the first implementation of the refactored config building; the version with the original pipeline.merge_config() ordering (commit "[this version has no step-pars values](https://github.com/spacetelescope/stpipe/pull/57/commits/26b01bfafe9228615d01eac7c5782c5185e0ecca)") still gets the step-pars overwritten by the combined spec-defaults+explicit cmdline par setting, while the most recent commit with the merge_config order swapped clobbers all spec+cmdline settings with the values contained in the step-pars file. 

I believe the issue is still a depth problem - the flow in cmdline.py being modified is only been accessed while `step_class` and `config` are for the pipeline - as such, the step-pars haven't been loaded into a config yet. This only happens for a step inside the pipeline during the pipeline __init__ call, during which each step calls `from_config_section` and either 

a) overwrites both the spec and the steps.tweakreg.kernel_fwhm with the step-pars values

or

b) overwrites the step-pars with spec + cmdline definitions.

Because of the depth in loading step-pars configs while building a pipeline config, either the cmdline args need to make it all the way into pipline __init__ or we need to make a pipeline version of one of these methods to push the cmdline args into the config merges another way.

tapastro:
This update pushes any parameter specifications:

either through command line arguments like `--steps.tweakreg.minobj=10`, or through a python session `Pipeline.call(input, steps={"tweakreg":{"config_file": "my_pars-tweakregstep.asdf", "kernel_fwhm": 1.23456}})`

downstream to the Pipeline __init__, due to the problem that any step-specific pars files provided in a pipeline initialization aren't loaded as a configobj until the pipeline is instantiated and begins working through its step_defs to instantiate the steps inside.

The current method has broken every test it could possibly break in jwst/stpipe, and is still a WIP.

hbushouse (approved):
Looks OK to me. As long as it works properly.

hbushouse:
At some point we should probably also include an entry for this fix in the jwst change log, just to have a record of the bug in jwst pipeline usage being fixed.

stscieisenhamer (approved):

mcara:
This PR broke support for list arguments

PR #63, opened 2022-07-15
Update get_crds_reference to not use JWST specific calls

nden:
The issue with the big memory increase was tracked down to how `ModelContainer.get_crds_parameters` was implemented. Every time it was called, it reopened the first model read in from the asn table, and called `get_crds_parameters` on it,  instead of calling `get_crds_parameters` directly on the already opened model. 

This means the fix should be in jwst.datamodels.container, however since the next jwst release is after the next romancal release I am implementing it here. I'll make the changes in jwst as well. After the next jwst release the code here can be simplified by just calling `model.get_crds_parameters` regardless of the model type.

Fixes AL-668

### DIFF
--- a/src/stpipe/pipeline.py
+++ b/src/stpipe/pipeline.py
@@ -181,7 +181,7 @@ class Pipeline(Step):
         # Iterate over the steps in the pipeline
         with cls._datamodels_open(dataset, asn_n_members=1) as model:
             if isinstance(model, Sequence):
-                crds_parameters = model._models.get_crds_parameters()
+                crds_parameters = model._models[0].get_crds_parameters()
                 crds_observatory = model.crds_observatory
             else:
                 crds_parameters = model.get_crds_parameters()
--- a/src/stpipe/pipeline.py
+++ b/src/stpipe/pipeline.py
@@ -182,24 +182,22 @@ class Pipeline(Step):
         with cls._datamodels_open(dataset, asn_n_members=1) as model:
             if isinstance(model, Sequence):
                 for contained_model in model:
-                    input_class = contained_model.__class__()
-                    metadata = input_class
-                    metadata.update(contained_model, only='PRIMARY')
+                    crds_parameters = model.get_crds_parameters()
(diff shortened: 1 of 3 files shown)


nden:
@tapastro Could you check this does no degrade performance for JP-2520 (the original issue).

tapastro:
This reverts back to initial behavior, i.e. ~25 GB of memory during prefetch.

src/stpipe/pipeline.py
```diff
@@ -181,25 +181,23 @@ def get_config_from_reference(cls, dataset, disable=None):
...
-                metadata = input_class
-                metadata.update(model, only='PRIMARY')
+                crds_parameters = model.get_crds_parameters()
+                crds_observatory = model.crds_observatory
 
         for cal_step in cls.step_defs.keys():
             cal_step_class = cls.step_defs[cal_step]
-            refcfg['steps'][cal_step] = cal_step_class.get_config_from_reference(metadata)
+            refcfg['steps'][cal_step] = cal_step_class.get_config_from_reference(crds_parameters,
+                                                                                 crds_observatory=crds_observatory)
```
tapastro, line 193: This line throws an error for me - seems as though crds_observatory isn't an input to this method. Removing it works just fine, though.

tapastro (changes requested):
With the one commented line removed, the behavior is good - memory usage is near zero during prefetch and I think is faster than it has ever been!

src/stpipe/pipeline.py
tapastro: Sorry, this comment was intended for L193 specifically - I removed `crds_observatory=crds_observatory` from the call and it runs without error.

nden: It was failing because `Pipeline.get_config_from_reference` also needed the `crds_observatory` parameter which I added in the latest commit.

nden:
@tapastro this is ready for review

tapastro (approved):
Looks good to me!

PR #67, opened 2022-07-28
Pull first model from ModelContainers before datamodel methods applied

tapastro:
Recent changes to config creation and reference file gathering require a fix for ModelContainer input, so datamodel.crds_parameters() will be run on the contents of the ModelContainer and not the container itself.


nden (approved):
Third time's a charm!

PR #68, opened 2022-07-28
Use ModelList marker instead of generic Sequence

tapastro:
Check instance of input model against ModelList rather than Sequence to differentiate behavior required for ModelContainer input versus typical datamodels.

DRAFT: requires PR approval in, and new build of, stdatamodels (expected to be 0.4.4).


stscieisenhamer (approved):
pending stdatamodels approval

stscieisenhamer:
This is still marked as draft. When ready, please rebase and un-draft.

tapastro:
Same story as jwst PR - no longer draft, but this needs to wait for a new stdatamodels release before merging.

braingram:
@tapastro I'm closing this PR as `stdatamodels` is no longer a dependency. Thanks for making this update and feel free to refresh and reopen this PR.

The use of `Sequence` does seem less-than-ideal as it matches `str` instances which I don't think is intended.

PR #75, opened 2022-12-15
refactor toxenvs, use OpenAstronomy workflows, and move MacOS jobs to schedule

zacharyburnett:
This PR abstracts CI workflows to the OpenAstronomy workflows; this should reduce maintenance for updating and maintaining actions. Additionally, these changes move the majority of MacOS jobs to the weekly scheduled workflow. This should alleviate the issue where the limited MacOS runners for the organization are not available for CI jobs.

- add `tox.ini` with `toxenv`s from Romancal
- test various Python versions on `ubuntu` and `macos`
- ~~use `ruff` instead of `flake8` to check lint rules~~ moved to #77

image

after merging, the Required Statuses should also be updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,20 +26,16 @@ jobs:
       envs: |
         - linux: test-xdist
           python-version: '3.8'
-        - macos: test-xdist
-          python-version: '3.8'
         - linux: test-xdist
           python-version: '3.9'
-        - macos: test-xdist
-          python-version: '3.9'
         - linux: test-xdist
-          python-version: '3.10'
-        - macos: test-xdist
           python-version: '3.10'
         - linux: test-xdist
           python-version: '3.11'
         - macos: test-xdist
           python-version: '3.11'
+        - linux: test-cov
+          coverage: 'codecov'
   test_numpy:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
     with:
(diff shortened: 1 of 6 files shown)


stscieisenhamer (commented):
General curmudgeon/grey-beard start about flake8 to ruff: Why?  Speed-up is a good thing, however version "0.0.187", though usually I do not take much stock in versions, still seems a bit early.  Big question is IDE support: Is there any? Also, losing the flake8 configuration will re-introduce unneeded noise. Also, how will this play with depending packages?

Regardless, will need some developer documentation at least stating what is being used.

zacharyburnett:
> General curmudgeon/grey-beard start about flake8 to ruff: Why? Speed-up is a good thing, however version "0.0.187", though usually I do not take much stock in versions, still seems a bit early. Big question is IDE support: Is there any? Also, losing the flake8 configuration will re-introduce unneeded noise. Also, how will this play with depending packages?
> 
> Regardless, will need some developer documentation at least stating what is being used.

Mainly, I am looking toward `ruff` for speedup, and support for `pyproject.toml` (to reduce clutter in the root directory). You're right however that it's still in development; I will separate the change into another PR so we can decide on it separately.

stscieisenhamer (approved):
LGTM.

Related things that could probably be done, in other PRs, include updating the README to document how to install, setup development, test, etc. About time this becomes a proper package.

Also, for the group itself, a discussion on the changes/new stuff, such as tox and linting changes, would be in order. Possibly at an all-hands.

WilliamJamieson:
@zacharyburnett it appears there is an issue with CRDS for jwst and romancal, can you take a look at these? I would like to merge this before merging #79.

tox.ini
```diff
@@ -16,12 +16,12 @@ envlist =
 #
 
 [testenv:check-style]
-description = check code style, e.g. with ruff
+description = check code style, e.g. with flake8
 skip_install = true
 deps =
-    ruff
+    flake8
```
WilliamJamieson, line 22: Why are you reverting the `ruff` change? PR #77 has been merged, so `ruff` is the style linter being used by `stpipe`.

tox.ini
```diff
@@ -53,6 +53,7 @@ extras =
     test
 deps =
     xdist: pytest-xdist
+    cov: pytest-cov
```
WilliamJamieson, line 56: Why is this an additional dependency, it is one of the `[test]` dependencies.

WilliamJamieson (changes requested):

tox.ini
zacharyburnett: I figure that this makes it explicit, so it doesn't necessarily need `[test]`. I can remove it if that doesn't make sense

WilliamJamieson:
@zacharyburnett it looks like CRDS is working for JWST now but not for Roman. Is this because there is not a public CRDS for Roman yet?

.github/workflows/ci.yml
```diff
@@ -12,120 +12,82 @@ on:
...
+        - linux: test-numpy120
+          python-version: '3.9'
+        - linux: test-numpy122
+          python-version: '3.8'
+        - linux: test-numpy121
+          python-version: '3.8'
+        - linux: test-numpy120
+          python-version: '3.8'
+        - linux: test-cov
+          coverage: 'codecov'
```
WilliamJamieson, line 56: I suggest breaking these into smaller sections. Like a `numpy` section or a `check` section. This will make browsing the CI easier to do from the CI overview.

WilliamJamieson: Something like the `asdf-astropy` CI: https://github.com/astropy/asdf-astropy/blob/main/.github/workflows/ci.yml

tox.ini
```diff
@@ -0,0 +1,88 @@
...
+    oldestdeps: minimum_dependencies stpipe --filename requirements-min.txt
+    oldestdeps: pip install -r requirements-min.txt
+    pip freeze
+commands =
+    pytest \
+    warnings: -W error \
+    xdist: -n auto \
+    jwst: --pyargs jwst --ignore-glob=timeconversion --ignore-glob=associations \
+    romancal: --pyargs romancal \
+    cov: --cov=src/stpipe --cov-config=pyproject.toml --cov-report=term-missing --cov-report=xml \
```
WilliamJamieson, line 72: the `*/scripts/*` in `jwst` are ignored for testing purposes by `jwst`. In stdatamodels, I had to add https://github.com/spacetelescope/stdatamodels/blob/514ff63dc4f357c197d1b7649b30817373e4d5a9/tox.ini#L54 in order to avoid the failures in these tests.

WilliamJamieson (changes requested):

pyproject.toml
```diff
@@ -12,7 +12,7 @@ classifiers = [
     'Programming Language :: Python :: 3',
 ]
 dependencies = [
-    'asdf>=2.8.0',
+    'asdf>=2.13',
```
WilliamJamieson, line 15: I think this should be announced in the changelog

WilliamJamieson (changes requested):

PR #76, opened 2022-12-16
fix `scripts` metadata to include `strun` as a standalone Python script in package

zacharyburnett:
by adding the `scripts` parameter to `setup()`, `strun` is included in the PATH when the package is installed


hbushouse (approved):
Looks OK to me, but then I don't really understand how it all works (that's why we have system/installation engineers).

PR #73, opened 2022-12-04
JP-3018: Restore support for step list arguments at command line

mcara:
Closes #69
Closes #71
Resolves [JP-3018](https://jira.stsci.edu/browse/JP-3018)

CC: @tapastro @hbushouse @stscieisenhamer @nden

This PR restores support for step's list arguments at the command line that was broken since #57 (or version `0.4.0`).

This PR essentially undoes most of the changes from #57 with exception of moving `_override_config_from_args()` in front of  `config = step_class.merge_config(config, config_file)` in `cmdline.just_the_step_from_cmdline()` - this was the only change compared to pre- #57 that was needed to fix the issue that #57 was addressing while not breaking support for list arguments.

-----

### Testing:

Here is my setup (use any MIRI imaging ASN as input data; non-MIRI might have different default values from CRDS):

For testing purposes, I created two step configuration files (for tweakreg and for the resample steps):

**tweakreg-custom.cfg**:
```INI
name = "tweakreg"
class = "jwst.tweakreg.TweakRegStep"

kernel_fwhm = 2.77
```

**resample-custom.cfg**:
```INI
name = "resample"
class = "jwst.resample.ResampleStep"

output_shape = 4900,4999
pixel_scale_ratio = 0.98765
```

I then run the pipeline like this:
```Shell
strun cfg/calwebb_image3.cfg miri_asn.json  --steps.resample.rotation=0 --steps.resample.output_shape=5000,5000
--steps.tweakreg.config_file=tweakreg-custom.cfg --steps.resample.config_file=resample-custom.cfg
```

Below I will reproduce only the relevant lines from the terminal output:
```
2022-12-04 11:18:56,323 - stpipe - INFO - PARS-TWEAKREGSTEP parameters found: /.../jwst/miri/jwst_miri_pars-tweakregstep_0019.asdf
2022-12-04 11:18:56,669 - stpipe.Image3Pipeline - INFO - Prefetch for DRIZPARS reference file is '/.../jwst/miri/jwst_miri_drizpars_0001.fits'.

2022-12-04 11:18:57,068 - stpipe.Image3Pipeline.tweakreg - INFO - Step tweakreg parameters are:
   {'pre_hooks': [], 'post_hooks': [], 'output_file': None, 'output_dir': None, 'output_ext': '.fits',
    'output_use_model': True, 'output_use_index': True, 'save_results': False, 'skip': False,
    'suffix': None, 'search_output_file': True, 'input_dir': '', 'save_catalogs': False,
[...]

PR #78, opened 2022-12-22
convert FromCommandLine to str before building Step

braingram:
When string arguments are passed in via the command line they are converted to FromCommandLine instances. Convert them back to strings before providing them as keyword arguments while building the step.

It appears this conversion is only needed to resolve the path of input and output files provided to the step
https://github.com/spacetelescope/stpipe/blob/454314bfd8cfd8764bfc55448bc7c1c2d0f6489f/src/stpipe/config_parser.py#L38-L46

This will hopefully fix the failing regression tests for https://github.com/spacetelescope/jwst/pull/7403 and other instances of similar failures.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.4.5 (unreleased)
 ==================
 
--
+- convert FromCommandLine instances to str before using as keyword arguments to Step [#78]
 
 0.4.4 (2022-12-16)
 ==================
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -260,12 +260,23 @@ class Step:
         if 'name' in config:
             del config['name']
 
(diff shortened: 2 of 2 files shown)


hbushouse:
Regtest run using this PR branch shows no errors for the niriss_soss tests that previously crashed!
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/518/
(other failures are expected, due to other mergers)

hbushouse (approved):
I don't know if it's the right change to make, but it seems to work!

nden:
This looks reasonable to me. I think the problem started when list arguments were enabled again in #73. 
Is it possible to come up with a test for this case?

nden:
I tested manually running a step with list parameters and confirmed this works. Writing a formal test can be done later.

nden (approved):

PR #79, opened 2023-01-06
Start using pre-commit

WilliamJamieson:
This PR adds a [pre-commit](https://pre-commit.com/) configuration and switches the `Code style checks` CI to using it instead of only running only `ruff`. The pre-commit includes additional useful checks. The checks added by this PR are:

- check-added-large-filesz: checks that "large" files are not added, which github might reject.
- check-ast: checks that all the python code will at least syntactically parse.
- check-case-conflict: checks that files will not conflict in case-insensitive file systems.
- check-yaml: checks that all yaml files are valid yaml.
- check-toml: checks that all toml files are valid toml.
- check-merge-conflict: checks that no merge-conflict statements have been left behind.
- check-symlinks: checks that all symlinks are valid and contained within the repository.
- debug-statements: checks that no python debugger statements are left in the code.
 - detect-private-key: checks that no private keys have been commit
- end-of-file-fixer: Properly fixes the end of files for the correct spacing (usually ending with empty line)
- trailing-whitespace: Checks that lines do not end with trailing whitespace (spaces, tabs, etc)
- python-check-blanket-noqa: checks that all `# noqa` include specific error codes to ignore.
- python-check-mock-methods: checks for common mistakes with mocks
- rst-directive-colons: checks that rst directives end with double colons
- rst-inline-touching-normal: checks that inline code code does not touch text in rst
- text-unicode-replacement-char: Forbits files with UTF-8 unicode replacement character.
- pyupgrade: automatically updates old python syntax to modern python syntax
- ruff: runs ruff on the code
- bandit: runs bandit security checs on the code.

Note that by providing a pre-commit configuration, developers can easily configure their local version of git to run all of the above checks (which are relatively fast) on any file changes prior to committing them. The majority of the above checks are automatically fixed by pre-commit if they fail and can then be subsequently re-staged, so that the commit succeeds. Alternately pre-commit can be used to run all the above checks with one command instead of several, while automatically fixing most errors.


zacharyburnett (approved):
Looks good, we could also take out the `check` matrix entirely with `pre-commit.ci`

WilliamJamieson:
> Looks good, we could also take out the `check` matrix entirely with `pre-commit.ci`

Almost, though I think we will need to do something about the `build-dist` job. OpenAstronomy has a good workflow that does a more through job of doing that test.

PR #80, opened 2023-01-07
Start using `isort` and `black` code formatters

WilliamJamieson:
Apply the the code formatters:

- [`isort`](https://pycqa.github.io/isort/) to consistently sort imports
- [`black`](https://black.readthedocs.io/en/stable/) to consistently format the code base.

These will help keep the codebase consistently formatted, and are recommend to be used in conjunction with the `ruff` linter that `stpipe` is already using.

I also adjusted the the line-length restriction from 130 down to a more reasonable (and the `black` standard) length of 88.

Note that this is base on PR #79 and so it should be merged after it.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,8 +2,7 @@
 
 [flake8]
 select = F, W, E101, E111, E112, E113, E401, E402, E501, E711, E722
-# We should set max line length to 88 eventually
-max-line-length = 130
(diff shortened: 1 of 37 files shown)


zacharyburnett (approved):
Looks good, nice commit separation

PR #81, opened 2023-01-09
Start using `codespell`

WilliamJamieson:
[`codespell`](https://github.com/codespell-project/codespell) is a tool which is designed to check for misspelled words in source code. This PR sets up a reasonable configuration for `codespell`, fixes all the errors `codespell` uncovered, and adds `codespell` to the basic style checks for `stpipe`. Note that `codespell` can both autofix many common misspellings and be configured to ignore some misspellings by adding words to the `ignore-words-list` in the `pyproject.toml` file.

Note that this PR is built on top of the changes in #80 and so should be merged after PR #80


zacharyburnett (approved):
This means I misspelled "precedence" across multiple repos haha

PR #91, opened 2023-04-11
Remove `stdatamodels` dependency

WilliamJamieson:
`romancal` has a dependency of `stdatamodels`; however, there should not be a dependency of the Roman pipeline on JWST pipeline specific packages.

The dependency is due to the `stdatamodels.s3_utils` module, which is a thing wrapper of the now unsupported `stsci-aws-utils` package. The introduction of `s3_utils` was largely due to an experiment to see if the `jwst` pipeline could load reference files from S3. While this experiment was apparently successful, it was intended to be a one-off demonstration and has been largely unused since it completed. @eslavich can provide more details. Moreover, the implantation of S3 support requires client code in `jwst` and `romancal` to know the source of input data (S3 or local filesystem) and handle the opening of internal files in a distinct fashion.  Ideally the steps consuming reference files should not have to be aware of the origin of their files. Instead, CRDS, which provides these files to the steps, should handle this type of detail.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,12 +3,12 @@
 
 - Remove use of deprecated ``pytest-openfiles`` ``pytest`` plugin. This has been replaced by
   catching ``ResourceWarning`` s. [#90]
-
 - Start using ``pre-commit`` to handle style checks. [#79]
 - Apply the ``isort`` and ``black`` code formatters and reduce the line length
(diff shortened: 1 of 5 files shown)


WilliamJamieson:
The JWST failures are expected, as the `s3_utils` related parts of it have not been removed.

nden (approved):

PR #93, opened 2023-04-14
drop support for Python 3.8

zacharyburnett:
Numpy is imminently dropping support for Python 3.8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,7 @@ jobs:
     with:
       envs: |
         - linux: test-oldestdeps-cov-xdist
-          python-version: 3.8
-        - linux: test-xdist
-          python-version: '3.8'
+          python-version: 3.9
         - linux: test-xdist
           python-version: '3.9'
         - linux: test-xdist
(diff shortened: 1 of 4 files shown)


braingram:
Similar to stcal roman (which supports 3.8) tests against the development version of stpipe.
https://github.com/spacetelescope/stcal/pull/162#issuecomment-1508884573

.readthedocs.yaml
```diff
@@ -16,9 +16,13 @@ formats:
   - htmlzip
   - pdf
 
+build:
```
WilliamJamieson, line 19: Can you convert this to using mamba like astropy uses.

WilliamJamieson (changes requested):

.readthedocs.yaml
zacharyburnett: Sure, but that's probably for another PR as it would also require an environment file

WilliamJamieson: Well the environment you targeted in this PR is causing the docs to fail.

zacharyburnett:
blocked by #94

WilliamJamieson (approved):

PR #94, opened 2023-04-14
Switch to using mamba for RTD

WilliamJamieson:
Properly switches RTD to using `mamba` to build python environment.


zacharyburnett (approved):
ah I understand now, not replacing the `docs` extra

WilliamJamieson:
This touches no code, only docs env so jwst tests don't need to pass.

PR #103, opened 2023-08-29
Update the style checks for stpipe

WilliamJamieson:
Cleanup and implement better style checks for stpipe


zacharyburnett (approved):
looks good to me

PR #107, opened 2023-09-05
deprecate `preserve_comments` fix spec parsing for inline comments with closing parenthesis

### LINKED ISSUE #105: Step.spec parsing fails when comment contains a closing parthesis
This issue came up while adding a spec to jwst ami analyze in:
https://github.com/spacetelescope/jwst/pull/7862

An attempt to add an inline comment to a spec entry like the following
```
src = string(default='A0V') # Source spectral type for model (Phoenix models)
```

Failed with an odd error:
```
ValidationError: Config parameter 'src': missing
```

Removing the parentheses enclosing "Phoenix models" fixed the issue.

This appears to be related to the `_inspec` argument provided to the `ConfigObj`:
https://github.com/spacetelescope/stpipe/blob/e82a1f09697dcf8d1b3de85fdae67d2feebd04fc/src/stpipe/config_parser.py#L202
Removing this argument allows the spec to be parsed with seemingly no errors. The spec/ConfigObj also still contains a `inline_comments` attribute with accurate values for the comments. I'm not sure what `preserve_comments` is supposed to be doing.

Perhaps one way to address this would be to:
- add a test for a spec with a comment that contains a closing parenthesis
- remove the `_inspec` argument
- deprecate `preserve_comments`

braingram:
Fixes #105 

The `preserve_comments` argument ends up mapping to the `_inspec` argument for `ConfigObj`. Setting this appears to break parsing of spec lines that contain inline comments with a closing parenthesis. The `preserve_comments` argument also appears to have no effect on parsing comments (aside from introducing the bug above).

This PR deprecates `preserve_comments` and removes internal usage of the `_inspec` argument. This should fix parsing inline comments for closing parenthesis (this is tested in the included test) and have no effect on comment parsing (also included in the test).

Doing a quick search through romancal and jwst I see no usage of the `preserve_comments` argument so I don't expect either of those packages to require updates. Hopefully regression tests (to be run below) will reveal if there are any issues with these changes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 - Print out ``jwst`` or ``romancal`` versions from ``strun --version``. [#98]
 - Print default parameter values for ``strun <step_alias> --help`` [#101]
 - Move ``strun`` to entrypoints [#101]
+- Deprecate ``preserve_comments`` fix spec parsing for inline comments with
+  a closing parenthesis [#107]
 
 0.5.0 (2023-04-19)
 ==================
--- a/src/stpipe/cmdline.py
+++ b/src/stpipe/cmdline.py
@@ -279,7 +279,7 @@ def just_the_step_from_cmdline(args, cls=None):
     # all of the expected reference files
 
     # load_spec_file is a method of both Step and Pipeline
-    spec = step_class.load_spec_file(preserve_comments=True)
+    spec = step_class.load_spec_file()
 
     parser2 = _build_arg_parser_from_spec(spec, step_class, parent=parser1)
 
--- a/src/stpipe/config_parser.py
+++ b/src/stpipe/config_parser.py
@@ -5,6 +5,7 @@ import logging
 import os
 import os.path
 import textwrap
+import warnings
 from inspect import isclass
 
 from asdf import ValidationError as AsdfValidationError
@@ -20,6 +21,7 @@ from .extern.configobj.configobj import (
     get_extra_values,
 )
 from .extern.configobj.validate import ValidateError, Validator, VdtTypeError
+from .utilities import _not_set
 
 # Configure logger
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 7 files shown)


braingram:
docs failure should be fixed with: https://github.com/spacetelescope/stpipe/pull/109

hbushouse (commented):
Looks OK to me. I'm just wondering if removing `preserve_comments` will have any effect on what you get when just listing the params for a given step or pipeline. For example, right now executing `strun -h jump` gives you the big list of outputs:
```
  --min_sat_area        minimum required area for the central saturation of snowballs
  --min_jump_area       minimum area to trigger large events processing
  --expand_factor       The expansion factor for the enclosing circles or ellipses
  --use_ellipses        deprecated
  --sat_required_snowball 
                        Require the center of snowballs to be saturated
  --min_sat_radius_extend 
                        The min radius of the sat core to trigger the extension of the core
  --sat_expand          Number of pixels to add to the radius of the saturated core of snowballs
  --edge_size           Size of region on the edges of NIR detectors where a sat core is not required
  --find_showers        Turn on shower flagging for MIRI
  --extend_snr_threshold 
                        The SNR minimum for detection of extended showers in MIRI
  --extend_min_area     Min area of emission after convolution for the detection of showers
  --extend_inner_radius 
                        Inner radius of the ring_2D_kernel used for convolution
  --extend_outer_radius 
                        Outer radius of the ring_2D_Kernel used for convolution
```
where the comments on each param are clearly displayed. Will those go away with this change?

braingram:
> strun -h jump

Thanks for taking a look and the questions.

The argument parsing and help output appears unchanged with this PR. Running `strun -h jump` using stpipe installed from this PR I see output that is identical to the output when run with stpipe main branch.

The spec is parsed using `ConfigObj` which stores inline comments in the `inline_comments` attribute. That attribute is used when constructing the argument parser (and help output) here:
https://github.com/spacetelescope/stpipe/blob/1ade40811376a43a3992080b5ba38f25fabc51b5/src/stpipe/cmdline.py#L90

hbushouse (approved):

braingram:
Thanks @WilliamJamieson and @hbushouse for the reviews.

PR #108, opened 2023-09-05
replace regex in configobj

### LINKED ISSUE #106: Bundled `configobj` subject to CVE-2023-26112
configobj is subject to a ReDoS: https://github.com/advisories/GHSA-c33w-24p9-8m24

This regex used to parse the config/spec items suffers from catastrophic backtracking:
https://github.com/spacetelescope/stpipe/blob/e82a1f09697dcf8d1b3de85fdae67d2feebd04fc/src/stpipe/extern/configobj/validate.py#L540

This is one useful write-up on the issue: https://www.regular-expressions.info/catastrophic.html

The upstream/bundled project appears abandoned but does have an open PR that appears to fix the offending regex:
https://github.com/DiffSK/configobj/pull/236

Given the level of control necessary to exploit the issue (it seems likely a `Step.spec` could be crafted to exploit it) it seems like a low risk as the step could also just include a `while True`. I haven't thought through what this means for command line options or if a ReDoS in python is limited to one cpu.

braingram:
fixes #106 

This is a contrived example (which ultimately fails with `stpipe.config_parser.ValidationError: Config parameter 'foo': missing`):
```python
﻿﻿import stpipe

n = 65000

class Foo(stpipe.Step):
    spec = "foo = " + ")" * n + "(" * n

f = Foo()
```

Running the above example with stpipe main takes ~12 seconds to fail. With this PR it takes ~0.7 seconds (on my machine).

As mentioned in #106 this seems like a low risk as the user could also write a step that includes a `while True`. It seems worthwhile to fix the regex as the replacement appears equivalent and faster.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,7 @@
 - Move ``strun`` to entrypoints [#101]
 - Deprecate ``preserve_comments`` fix spec parsing for inline comments with
   a closing parenthesis [#107]
+- Fix regex issue in internal configobj [#108]
 
 0.5.0 (2023-04-19)
 ==================
--- a/src/stpipe/extern/configobj/validate.py
+++ b/src/stpipe/extern/configobj/validate.py
@@ -537,7 +537,7 @@ class Validator:
     """
 
     # this regex does the initial parsing of the checks
-    _func_re = re.compile(r'(.+?)\((.*)\)', re.DOTALL)
+    _func_re = re.compile(r'([^\(\)]+?)\((.*)\)', re.DOTALL)
 
     # this regex takes apart keyword arguments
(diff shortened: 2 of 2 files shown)


braingram:
Roman regtest shows the usual random error due to reference file mishandling:
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/340/pipeline/206
The above should be addressed with: https://github.com/spacetelescope/romancal/pull/858

JWST regtest all tests passed but some odd failure occurred at the end of test runs (which looks due to xdist not tearing down cleanly).
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/906/

Given the low impact of the issue addressed in this PR I'm going to leave it as draft until the above romancal PR is merged to hopefully improve the reliability of the regtests.

WilliamJamieson (approved):

PR #109, opened 2023-09-06
remove unsupported system_packages rtd config setting

braingram:
RTD no longer supports any system_packages setting:
https://blog.readthedocs.com/drop-support-system-packages/

This PR removes the unused system_packages setting

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -22,7 +22,6 @@ sphinx:
 # Install regular dependencies.
 # Then, install special pinning for RTD.
 python:
-  system_packages: false
   install:
     - method: pip
       path: .


zacharyburnett (approved):

zacharyburnett:
good catch! I'll apply this to other repos as well

braingram:
> good catch! I'll apply this to other repos as well

Thanks!

stdatamodels (https://github.com/spacetelescope/stdatamodels) and jwst (https://github.com/spacetelescope/jwst/pull/7860) should be all set.

@WilliamJamieson has an approved PR for roman_datamodels: https://github.com/spacetelescope/roman_datamodels/pull/266
and a merged one for romancal: https://github.com/spacetelescope/romancal/pull/853

rad doesn't appear to be effected: https://github.com/spacetelescope/rad/blob/main/.readthedocs.yml
stcal does: https://github.com/spacetelescope/stcal/blob/eba3b6a638e78c5d100bdf2da4dc1f787cfe05a2/.readthedocs.yaml#L25
beyond that I'm not sure.

PR #111, opened 2023-09-19
use PyPI upload workflow from OpenAstronomy

zacharyburnett:
This PR changes the PyPI upload process to use the OpenAstronomy workflow instead of the STScI one. The workflow in this PR is configured to test the build with each pull request push, and then publish upon an official release.

~Additionally, we can't use this workflow until we make a PyPI token (with write / upload permission to packages) for the STScI maintainer account. Currently, we use a username and password to PyPI; however, this method will be deprecated at some point in the future in favor of tokens.~


.github/workflows/publish-to-pypi.yml
```diff
@@ -2,15 +2,11 @@ name: Publish to PyPI
 
 on:
   release:
-    types: [released]
+    types: [ released ]
```
WilliamJamieson, line 5: Why limit this to just releasing? Really this should be run (without uploading) on every PR and be manually triggerable.
```suggestion
    types: [released, pull_request, workflow_dispatch]
```

.github/workflows/publish-to-pypi.yml
```diff
@@ -2,15 +2,11 @@ name: Publish to PyPI
...
 
 jobs:
   publish:
-    uses: spacetelescope/action-publish_to_pypi/.github/workflows/workflow.yml@master
-    with:
-      test: false
-      build_platform_wheels: false # Set to true if your package contains a C extension
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
```
WilliamJamieson, line 9: To not upload to PyPi for non-releases:
```suggestion
    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
    with:
      upload_to_pypi: ${{ gitub.event_name == 'released'}}
```
Then the upload will only be triggered when the action is triggered by releases.

WilliamJamieson (commented):
With my suggestions below, https://github.com/spacetelescope/stpipe/blob/510e4a24170faa0345e165625f13520d79e22d88/.github/workflows/ci.yml#L26 can be removed as the workflow should be triggered by pull_requests, but not be uploaded to PyPi.

.github/workflows/publish-to-pypi.yml
zacharyburnett: wow, powerful options!

.github/workflows/ci.yml
```diff
@@ -23,7 +23,6 @@ jobs:
     with:
       envs: |
         - linux: check-style
-        - linux: build-dist
```
WilliamJamieson, line 26: Now that I look at it, pre-commit is doing the check-style too, so the whole `check` block can be removed.

zacharyburnett:
> I wonder why the publish workflow has not been triggered... Would a close/reopen work to trigger that?

found it, it was just a syntax error

zacharyburnett:
unfortunately, it looks like GitHub categorizes the actions on the Actions tab by filename, so we can't rename the actual workflow files without creating a new category with the same name. If people see `Publish to PyPI` running on their pull request, they might panic and think it's actually publishing; thus I renamed the job to `build`, but am unable to rename `publish-to-pypi.yml`

.github/workflows/build.yml
```diff
@@ -0,0 +1,15 @@
+name: build
```
WilliamJamieson, line 1:
```suggestion
name: build wheels
```

.github/workflows/test.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: test
```
WilliamJamieson, line 1:
```suggestion
name: CI
```

.github/workflows/test_on_schedule.yml
```diff
@@ -1,5 +1,4 @@
-
-name: Weekly cron
+name: test on schedule
```
WilliamJamieson, line 1:
```suggestion
name: Weekly CI
```

PR #116, opened 2023-10-10
use `pytest.importorskip` on test utils

zacharyburnett:
closes #114

use `pytest.importorskip` to import util to make test image, to keep abstract model tests from failing if `roman_datamodels` is not installed

https://github.com/spacetelescope/stpipe/blob/58def5a69834835e5d9b2cd7ad97bbef36c24d8c/src/stpipe/tests/test_abstract_datamodel.py#L8-L13

PR #118, opened 2023-10-20
Add AbstractModelContainer

braingram:
This PR adds `AbstractModelContainer` which behaves similarly to [AbstractDataModel](https://github.com/spacetelescope/stpipe/blob/master/src/stpipe/datamodel.py).

The interface/protocol for `AbstractModelContainer` was constructed to:
- comply with [romancal.datamodel.container.ModelContainer](https://github.com/spacetelescope/romancal/blob/c5864bf4dac9bb8f2e66aba90da98e3e8fdf2414/romancal/datamodels/container.py#L30)
- comply with [jwst.datamodels.container.ModelContainer](https://github.com/spacetelescope/jwst/blob/6cee5c1abe5f7ed5ba4fc5d6b81612829bfdc61e/jwst/datamodels/container.py#L30)
- Include API required by stpipe:
  - have a `crds_observatory` attribute
  - be iterable (have an `__iter__` attribute)
  - have a `save` method
- and to include a few more common requirements that also allow this abstract class to be distinct from `AbstractDataModel`
  - have a `read_asn` method
  - have a `from_asn` method

To test the new `AbstractModelContainer` the jwst and romancal tests were updated. This conflicts with the changes in https://github.com/spacetelescope/stpipe/pull/116 It expands upon the changes in that PR to include running the integration tests when jwst and romancal/roman_datamodels are installed in the CI and extends those tests to test both `AbstractDataModel` and `AbstractModelContainer`.

This PR is likely an alternative to https://github.com/spacetelescope/stdatamodels/pull/102


braingram:
Roman regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/433/
JWST regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1009/

braingram:
romancal tests failing due to missing webbpsf data in the ci for stpipe, the first 2 tests run (the ones from `test_roman.py` added in this PR) appear to have passed.

braingram:
@schlafly I don't seem to be able to request you as a reviewer for this PR. Given that it relates to roman (and jwst) is there someone that can review this with an eye for the romancal interaction?

WilliamJamieson:
> This PR adds `AbstractModelContainer` which behaves similarly to [AbstractDataModel](https://github.com/spacetelescope/stpipe/blob/master/src/stpipe/datamodel.py?rgh-link-date=2023-10-20T17%3A18%3A41Z).

I do not like this approach as it digs us further into reliance on old difficult to follow Python practices. Instead this (and for that matter `AbstractDataModel`) should be defined using [`typing.runtime_checkable`](https://docs.python.org/3/library/typing.html#typing.runtime_checkable) Python [`typeing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) generic objects. Essentially, Protocols (which the changes suggested here are not actually a protocol in the python sense) are a way for library authors to consistently define a template for what an object which contains expected features for the library to hook into.

This is precisely what the intent for `AbstractDataModel` is attempting to provide. Indeed, https://github.com/spacetelescope/stpipe/blob/de7afd95904168162c7047c54ca81a7b9f056f5c/src/stpipe/datamodel.py#L18-L31
makes it so that any object containing `crds_observatory`, `get_crds_parameters`, and `save` attributes will pass an `isinstance(object, AbstractDataModel)` check without any need for inheritance. The lack of inheritance is necessary because there is no good reason for "DataModel-libraries" which are "`stpipe`-compatible" to require `stpipe` as a dependency. In fact, this is counter-productive in general as those libraries should be functional outside of pipeline execution code, e.g. opening the results of a pipeline run. For example both:

- [`stdatamodels.DataModel`](https://github.com/spacetelescope/stdatamodels/blob/ed1c5f6ee6d361b6a0b564af66b766b661cb704d/src/stdatamodels/model_base.py#L52) for JWST
- and [`roman_datamodels.DataModel`](https://github.com/spacetelescope/roman_datamodels/blob/33368d17f747d098858491e2121edb07f2ed48ac/src/roman_datamodels/datamodels/_core.py#L51) for ROMAN

do NOT have any inheritance from `AbstractDataModel` despite being used in our pipelines as `AbstractDataModel` objects when passed to `stpipe`. Indeed, to my knowledge there is no actual code which inherits from `AbstractDataModel` instead all usages of `AbstractDataModel` are really treatments of it as if it was a Python `Protocol`, which it is not.

Since `AbstractDataModel` is not in fact a true Python `Protocol`, things like modern IDE tooling don't correctly flag/handle things surrounding its effective use quite correctly. It also is quite repetitive in its construction as it attempts to act as a `Protocol` and an abstract base class, despite there being no intention of using it as a base class for anything.

Thus I suggest that the proposed notion of an `AbstractModelContainer` be dropped in favor of simply defining a `ModelContainer` Protocol instead.

schlafly:
Sorry, I can't provide useful feedback about this kind of thing.  Since Roman & JWST regression tests pass it looks fine to me and I would approve were I able.  I will defer to you and William on protocols vs. abstract interfaces vs. ...

braingram:
I'm closing this in favor of changes to the `ModelContainer` api may help to address memory issues in JWST.

PR #120, opened 2023-10-23
Switch abstract interfaces to protocols

WilliamJamieson:
This PR replaces the `AbstractDataModel` object with a python protocol as described in https://github.com/spacetelescope/stpipe/pull/118#issuecomment-1775692988. This is a bit cleaner than the current implementation.

~This PR also implements a `ModelContainer` protocol which provides the same functionality of #118.~

Note that the only functional change is that `issubclass` will does not function quite correctly due to the missing `__hash__` see https://github.com/python/mypy/issues/3939 for a more detailed discussion of this. Since in general the idea of the `AbstractDataModel` was to provide a way to check if some non-inheriting object instance followed the structure it described, `issubclass` checks are unnecessary and likely only limited to `stpipe` itself, which itself makes no such checks.


WilliamJamieson:
Passes regression tests:

- https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1016/
- https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/446/ (the one failure is known)

perrygreenfield (commented):
LGTM aside from one docstring that looks like it was mistakenly copied from elsewhere

PR #133, opened 2024-01-16
Fix pre- and post-hooks

jdavies-st:
Currently pre- and post-hooks only work as strings pointing to importable Step subclasses.  This made sense back when the foreseen use case of the pipeline was with `strun` at the command line, where all args are strings.

This PR fixes hooks to work with our current use cases.  It adds the possibility of using a defined function or defined `Step` subclass as hook, or an instance with parameters set of a `Step` subclass.  And it tests all the cases.

So now a hook can be a list of a number of useful things, not just a list of strings.

Hooks before:

```python
post_hooks = ["mypackage.MyFirstStep", "mypackage.MySecondStep"]  # string pointing to Step subclass
post_hooks = ["mypackage.my_function"]  # string pointing to a function that is Step.process() equivalent
post_hooks = ["fitsinfo {0}"]  # string pointing to system call
```

With this PR, we can also now use the following as hooks:

```python
from mypackage import MyFirstStep, my_function


post_hooks = [MyFirstStep]  # imported Step subclass
post_hooks = [my_function]  # imported function that is Step.process() equivalent
post_hooks = [MyFirstStep(param1=True, param2=3.5)]  # imported Step instance
post_hooks = ["mypackage.MyFirstStep(param1=True, param2=3.5)"]  # string pointing to Step subclass instance
```

Also added an extra CI job that runs the unit tests with some downstream dependencies.  There's always been tests in this package that depend on `stdatamodels` and `roman_datamodels`, but these have been skipped since the recent-ish github actions workflow updates, and subsequently the `roman_datamodels` test had a broken import.  I've chosen to test the hooks with `jwst` as an optional test dependency, as that's the lowest overhead way to do it without having to create a bunch of redundant test infrastructure.

### DIFF
--- a/src/stpipe/hooks.py
+++ b/src/stpipe/hooks.py
@@ -1,6 +1,7 @@
 """
 Pre- and post-hooks
 """
+import ast
 import inspect
 
 from . import function_wrapper, utilities
@@ -40,6 +41,18 @@ def hook_from_string_or_class(step, hooktype, num, command):
         except ImportError:
             # String is possibly a subproc, so handle this later
             pass
+        except AttributeError:
+            # String points to an instance of a Step
+            # So import the class
+            class_string, _, params = command.partition("(")
+            step_class = utilities.import_class(
+                class_string, subclassof=Step, config_file=step.config_file
+            )
+            # Then convert rest of string to args and instantiate the class
+            kwargs_string = params.strip(")")
+            expr = ast.parse(f"dict({kwargs_string}\n)", mode="eval")
+            kwargs = {kw.arg: ast.literal_eval(kw.value) for kw in expr.body.keywords}
+            return step_class(**kwargs)
         except TypeError:
             # String points to a function
             step_func = utilities.import_func(command)
--- a/src/stpipe/hooks.py
+++ b/src/stpipe/hooks.py
@@ -8,9 +8,9 @@ from . import function_wrapper, utilities
 from .step import Step
 
 
-def hook_from_string_or_class(step, hooktype, num, command):
+def hook_from_string(step, hooktype, num, command):
     """
-    Generate hook from string (or pass along the class)
+    Generate hook from string, function, Step or Step instance
 
     Parameters
     ----------
@@ -105,7 +105,4 @@ def get_hook_objects(step, hooktype, hooks):
     -------
     list of callables that can be run as a hook
     """
-    return [
-        hook_from_string_or_class(step, hooktype, i, hook)
-        for i, hook in enumerate(hooks)
-    ]
+    return [hook_from_string(step, hooktype, i, hook) for i, hook in enumerate(hooks)]
--- a/src/stpipe/hooks.py
+++ b/src/stpipe/hooks.py
@@ -1,15 +1,14 @@
 """
 Pre- and post-hooks
 """
-import contextlib
 import inspect
-import types
 
+from . import function_wrapper
 from . import utilities
 from .step import Step
 
 
-def hook_from_string_or_class(step, type, num, command):
+def hook_from_string_or_class(step, hooktype, num, command):
     """
     Generate hook from string (or pass along the class)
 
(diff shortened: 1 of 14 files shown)


jdavies-st:
@hbushouse this is the promised PR to make pre- and post-hooks work within python instead of just with the `strun` CLI.  There's still some tests in `jwst` that use .cfg files for hooks.  Once this is merged and released, we can update those properly.

The only use case I have not thought about is if one wanted to do something like:

```
$ strun calwebb_detector1 jw001234_blah_blah_uncal.fits --steps.jump.post_hooks="mypackage.MyAwesomeStep(threshold=3.5)",
```

i.e. passing an instance of a step via the CLI as a hook.  One can pass an instance of a step as a hook within python, so perhaps this is not worth pursuing.

braingram:
JWST regtests running at: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1148/
roman regtests running at: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/551/

jdavies-st:
@drlaw1558 could you show the cells of your code?

If you define a step in one cell:

```python
class XplyStep(Step):

    class_alias = "xply"

    def process(self, input_data):
        pass
```

Then the way to use it later as a hook, is directly:


```python
steps = dict(
    assign_wcs = dict(
        post_hooks = [
            XplyStep,
        ],
    ),
)
result = Spec2Pipeline.call(input_data, steps=steps)
```

No need to pass it as `xply.XplyStep`, as that's only if its in a module `xply` that is importable.

drlaw1558:
Ah, that was it.  I'd removed the xply., but I also needed to remove the quotes.  I.e.,

dict = {"assign_wcs": {"post_hooks": ["XplyStep"]}}
but needed
dict = {"assign_wcs": {"post_hooks": [XplyStep]}}

Looks good to me then, and the logging functionality is now working too.

jdavies-st:
Any chance of this getting a review?  From a user perspective, it would be really nice to have this improved `hook` functionality in the next release of `jwst`.

src/stpipe/hooks.py
```diff
@@ -1,40 +1,111 @@
...
-import types
+import ast
+import inspect
 
-from . import utilities
+from . import function_wrapper, utilities
 from .step import Step
 
 
-def hook_from_string(step, type, num, command):  # noqa: A002
```
nden, line 11: Should this be deprecated instead of removed? It looks like it's been available for a long time.

nden:
@jdavies-st I think this looks good - other than the one comment I have.

src/stpipe/hooks.py
jdavies-st: It's not removed, I've renamed the `hook_from_string()` function to `hook_from_string_or_class()` to better describe what it actually does.  Happy to change the name back if you think this will affect the public API.  Since this is all internal workings, I would rather make all this private, including the whole `hooks.py` module, but happy to maintain backwards compat.

If not maintaining backwards compat, maybe "hook_from_object()` would be a better function name.

`stpipe` suffers from a lot of its internal workings, such as `Step.run()`, being public functions.

As far as I can tell, a search of Github doesn't turn up any uses of `stpipe.hooks` outside of `stpipe`.

jdavies-st: I've gone ahead and changed the function name back to its original name, and just clarified in the docstring what it does.

jdavies-st:
Once this is merged and stpipe 0.6.0 is released, I'll go ahead and update the docs in `jwst` to reflect the new possibilities of `hook` usage.

nden:
Regression tests pass again.

src/stpipe/hooks.py
nden: Thanks! And I agree it's unlikely anyone uses it.

nden (approved):

nden:
@jdavies-st Please ping me on the JWST PR.

PR #134, opened 2024-01-18
Fix datetime.utcnow() deprecation

jdavies-st:
This fixes the following deprecation warning:

```
DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC)
```

Note that when you pass any timezone to `datetime.now()`, it returns the offset from UTC (format +HH:MM) appended to the string.  The offset from UTC is always zero, and `datetime.utcnow()` didn't include it.  So here it is removed so that timestamps have a consistent format.

### DIFF
--- a/src/stpipe/config.py
+++ b/src/stpipe/config.py
@@ -4,7 +4,7 @@ will eventually fully replace config_parser.py, but we'll need to maintain
 both until we replace configobj with traitlets.
 """
 from copy import deepcopy
-from datetime import UTC, datetime
+from datetime import datetime, timezone
 
 import asdf
 
@@ -111,7 +111,7 @@ class StepConfig:
             meta = deepcopy(_META_TEMPLATE)
             meta["date"] = meta["date"].replace(
                 _TEMPLATE_PLACEHOLDER,
-                datetime.now(UTC)
+                datetime.now(timezone.utc)
                 .replace(microsecond=0)
                 .isoformat()
                 .removesuffix("+00:00"),
--- a/src/stpipe/config.py
+++ b/src/stpipe/config.py
@@ -4,7 +4,7 @@ will eventually fully replace config_parser.py, but we'll need to maintain
 both until we replace configobj with traitlets.
 """
 from copy import deepcopy
-from datetime import datetime
+from datetime import UTC, datetime
 
 import asdf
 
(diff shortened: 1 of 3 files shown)


src/stpipe/config.py
```diff
@@ -4,7 +4,7 @@
 both until we replace configobj with traitlets.
 """
 from copy import deepcopy
-from datetime import datetime
+from datetime import UTC, datetime
```
braingram, line 7:
```suggestion
from datetime import datetime, timezone
```

src/stpipe/config.py
```diff
@@ -111,7 +111,10 @@ def to_asdf(self, include_metadata=False):
             meta = deepcopy(_META_TEMPLATE)
             meta["date"] = meta["date"].replace(
                 _TEMPLATE_PLACEHOLDER,
-                datetime.utcnow().replace(microsecond=0).isoformat(),
+                datetime.now(UTC)
```
braingram, line 114:
```suggestion
                datetime.now(timezone.utc)
```
I believe this and the other suggestion should allow the CI to work with 3.9 and matches the [changes made in jwst](https://github.com/spacetelescope/jwst/pull/8051/files#diff-d0a38e200cb582a0443f0e448a2eb55c044ab229869df889ed7e5c633facb8b5R567).

Although not necessary with this PR, would you add a python 3.12 environment to tox and the CI? This would be great to verify that the `DeprecationWarning` is fixed.

braingram (commented):
Thanks for making this update! I made a suggestion that should work with python 3.9 to fix the CI.

src/stpipe/config.py
jdavies-st: Thanks.  Done.

This actually doesn't get picked up in the `stpipe` unit tests, but does in the `jwst` unit tests, which test `config_parser.py` better than `stpipe` does.  One might want to think about adding coverage to the downstream tests, specifically `jwst`.

braingram (approved):
Thanks! I totally missed that the lines weren't covered by the unit tests. It might be worth moving some of those over (if that's possible) but that's a problem for another day.

LGTM.

jdavies-st:
Agree that it would be nice if this package could self-test better.  Another day.

I will note that by putting coverage on the downstream test workflows and adding codecov flags, you can keep track separately of which coverage is coming from the package unit tests and how well the downstream packages test it.  But currently the OpenAstronomy CI workflows don't support specifying codecov flags, though it would be easy enough to add probably.  Flags are used in the `jwst` package Jenkins regtests currently, though the unit ones disappear with the switch to OpenAstronomy CI workflows.

PR #143, opened 2024-02-26
Fix handling of pathlib.Path objects

### LINKED ISSUE #139: Support for `pathlib.Path` objects?
Hi!

I was wondering if `pathlib.Path` objects were supported in `stpipe` (and packages depending on it like `jwst`) or if support was planned in the future.

I wanted to report an example issue I encountered recently: using a `Path` object as input to `Step.run()` leads to the `_input_filename` attribute not being set because the `Path` falls in the `else` category [here](https://github.com/spacetelescope/stpipe/blob/2a976365406b11ae6f116fe0e97f529fee700b9c/src/stpipe/step.py?plain=1#L893-L902).

The simple fix for users is to use strings as input path, but it took me a bit of time to figure this out because I missed the error message in the log.

jdavies-st:
This fixes handling of `pathlib.Path` objects when used for input files for `Step`.

It also removes the old, deprecated `tmpdir` pytest fixture and replaces it with the modern `tmp_path` equivalent, which provides a `pathlib.Path` instance.

Resolves #139

### DIFF
--- a/src/stpipe/log.py
+++ b/src/stpipe/log.py
@@ -5,6 +5,7 @@ import fnmatch
 import io
 import logging
 import os
+import pathlib
 import sys
 import threading
 from contextlib import contextmanager
@@ -176,7 +177,7 @@ def load_configuration(config_file):
 
     Parameters
     ----------
-    config_file : str or readable file-like object
+    config_file : str, pathlib.Path instance or readable file-like object
     """
 
     def _level_check(value):
@@ -192,6 +193,8 @@ def load_configuration(config_file):
         return value
 
     spec = config_parser.load_spec_file(LogConfig)
+    if isinstance(config_file, pathlib.Path):
+        config_file = str(config_file)
     config = ConfigObj(config_file, raise_errors=True, interpolation=False)
     val = validate.Validator()
     val.functions["level"] = _level_check
--- a/tests/test_step.py
+++ b/tests/test_step.py
@@ -76,9 +76,9 @@ class ListArgStep(Step):
 
 
 @pytest.fixture()
-def config_file_pipe(tmpdir):
+def config_file_pipe(tmp_path):
     """Create a config file"""
-    config_file = str(tmpdir / "simple_pipe.asdf")
+    config_file = tmp_path / "simple_pipe.asdf"
 
     tree = {
         "class": "test_step.SimplePipe",
@@ -104,9 +104,9 @@ def config_file_pipe(tmpdir):
 
 
 @pytest.fixture()
-def config_file_step(tmpdir):
+def config_file_step(tmp_path):
     """Create a config file"""
-    config_file = str(tmpdir / "simple_step.asdf")
+    config_file = tmp_path / "simple_step.asdf"
 
     tree = {
         "class": "test_step.SimpleStep",
@@ -122,9 +122,9 @@ def config_file_step(tmpdir):
 
 
 @pytest.fixture()
-def config_file_list_arg_step(tmpdir):
+def config_file_list_arg_step(tmp_path):
     """Create a config file"""
-    config_file = str(tmpdir / "list_arg_step.asdf")
+    config_file = tmp_path / "list_arg_step.asdf"
 
     tree = {
         "class": "test_step.ListArgStep",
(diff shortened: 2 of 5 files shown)


tests/test_step.py
```diff
@@ -284,9 +284,9 @@ def test_step_list_args(config_file_list_arg_step):
     correctly.
     """
     config, returned_config_file = ListArgStep.build_config(
-        "science.fits", config_file=config_file_list_arg_step
+        "science.fits", config_file=str(config_file_list_arg_step)
```
braingram, line 287: As updating `build_config` to also accept `Path` seems to match the efforts in this PR would you update that portion of `Step` to also accept a `Path` instance:
https://github.com/spacetelescope/stpipe/blob/ee5f1ee8e505b43faaad9e86409391274407d8b5/src/stpipe/step.py#L1376-L1381
That should make this call to `str` unnecessary.

src/stpipe/log.py
```diff
@@ -192,6 +193,8 @@ def _level_check(value):
         return value
 
     spec = config_parser.load_spec_file(LogConfig)
+    if isinstance(config_file, pathlib.Path):
+        config_file = str(config_file)
     config = ConfigObj(config_file, raise_errors=True, interpolation=False)
```
braingram, line 198: What do you think of the following as an alternative to the `isinstance` conditional?
```suggestion
    config = ConfigObj(str(config_file), raise_errors=True, interpolation=False)
```

braingram:
Thanks for putting this together!

Do the changes in this PR allow us to use `-p no:legacypath`? (see https://github.com/asdf-format/asdf/pull/1759 for details). If so I'd be all for adding that to `pyprojec.toml` to guard against future tests using `tmpdir`.

tests/test_step.py
jdavies-st: This actually fails within `ConfigObj` unfortunately.  It does not fully accept `pathlib.Path` objects.  And since this is now a dependency instead of vendored in `extern`, not much I can do here.  Though perhaps I could modify line 1379 to pass a `str()` of the path?

jdavies-st:
> Do the changes in this PR allow us to use `-p no:legacypath`? (see [asdf-format/asdf#1759](https://github.com/asdf-format/asdf/pull/1759) for details). If so I'd be all for adding that to `pyprojec.toml` to guard against future tests using `tmpdir`.

Good question.  I'll give it a try. I too was trying to think of a way to prevent the use of the `tmpdir` pytest fixture from being reintroduced in testing as well.  👍

tests/test_step.py
braingram: It looks like `config_file` is returned by this function and is documented to be a `str`. So perhaps converting `config_file` to a `str` on 1377?

src/stpipe/log.py
braingram: Thanks for checking. It looks like I missed that `config_file` might also be a file-like object.

jdavies-st: Ah, and that's why in the test when we send it an `io.StringIO` instance, it barfs.  Thanks.  I'll keep the conditional here then.

Though honestly, I think passing around bytes objects in the logger is bit, um, old-fashioned.  I will leave this mess to future refactoring.  =)

jdavies-st: I will revert it in the branch.

pyproject.toml
```diff
@@ -71,7 +71,8 @@ addopts = [
     "--strict-config",
     "--strict-markers",
     "-ra",
-    "--color=yes"
+    "--color=yes",
+    "-p no:legacypath",
```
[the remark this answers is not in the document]
jdavies-st: Sounds good.  I'll remove this, and I'll open an issue here and in the downstream packages.

braingram:
Thanks for all your work on this!

Regression tests running at:
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/611/
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1231/

braingram (approved):
Roman regtests all passed
JWST regtests had 3 failures (all known and random failures for missing log messages that occur likely due to stpipe logger handling).

Issue #145, opened 2024-02-27
Use tmp_path instead of tmpdir pytest fixture


jdavies-st:
Replace all instances thereof, and then in pyproject.toml:

```toml
[tool.pytest.ini_options]
addopts = [
    "-p no:legacypath",
]
```

This is will make the code more robust for use of `pathlib.Path` instances in the runtime code.

PR #146, opened 2024-02-28
enable nolegacypath tox factor

braingram:
Fixes: https://github.com/spacetelescope/stpipe/issues/145

Add a `nolegacypath` tox factor that run pytest with `-p no:legacypath` which will produce errors if legacy "non-path" fixtures/features are used.

A tox factor was added to allow the downstream jobs (like romancal and jwst) to run without `-p no:legacypath` as we are mainly concerned here that stpipe does not use legacy path features in it's test suite (not the test suites of romancal and jwst).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
         - macos: py311-xdist
         - linux: py311-cov-xdist
           coverage: 'codecov'
-        - linux: py312-xdist
+        - linux: py312-xdist-nolegacypath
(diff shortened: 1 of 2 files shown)


jdavies-st (approved):
🎉

braingram:
Thanks @jdavies-st for the review. I just requested @nden for a review to hopefully appease the branch protection (I'm assuming @jdavies-st is not listed as an approved reviewer?).

nden (approved):

PR #153, opened 2024-05-28
WIP: JP-3563: Raise warning instead of error for unexpected values in config

melanieclarke:
Resolves [JP-3563](https://jira.stsci.edu/browse/JP-3563)

Closes https://github.com/spacetelescope/jwst/issues/8335

For future-proofing pipeline runs with new steps, unexpected values in pipeline or step configuration now raise warnings instead of ValidationErrors.  Unexpected steps or parameter values are ignored.  This allows an older pipeline version to be used with a newer parameter reference file.

### DIFF
--- a/tests/test_config_parser.py
+++ b/tests/test_config_parser.py
@@ -77,9 +77,7 @@ def test_validate_extra_value_warning(action, monkeypatch):
 
     spec = config_parser.load_spec_file(MockStep)
 
-    monkeypatch.setattr(
-        config_parser, "EXTRA_VALUE_WARNING_ACTION", action
-    )
+    monkeypatch.setattr(config_parser, "EXTRA_VALUE_WARNING_ACTION", action)
     if action == "error":
         # Error is raised
         with pytest.raises(
--- a/tests/test_config_parser.py
+++ b/tests/test_config_parser.py
@@ -61,8 +61,9 @@ def test_preserve_comments_deprecation(value):
     assert "inline comment (with parentheses)" in spec.inline_comments["bar"]
 
 
-@pytest.mark.parametrize("action",
-                         ["default", "error", "ignore", "always", "module", "once"])
+@pytest.mark.parametrize(
+    "action", ["default", "error", "ignore", "always", "module", "once"]
+)
 def test_validate_extra_value_warning(action):
     """Test that extra values in the configuration raise warnings.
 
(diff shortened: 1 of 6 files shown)


ddavis-stsci (approved):
LGTM

tests/test_config_parser.py
```diff
@@ -58,3 +59,38 @@ class Foo:
...
+    The warning behavior can be configured by modifying the
+    EXTRA_VALUE_WARNING_ACTION switch for the module
+    """
+    config = ConfigObj({"expected": True, "unexpected": False})
+
+    class MockStep:
+        spec = "expected = boolean(default=False) # Expected parameter"
+
+    spec = config_parser.load_spec_file(MockStep)
+    config_parser.EXTRA_VALUE_WARNING_ACTION = action
```
braingram, line 79: A [monkeypatch](https://docs.pytest.org/en/6.2.x/monkeypatch.html) would be useful here to reset `EXTRA_VALUE_WARNING_ACTION` back to the default value when the test finishes.

melanieclarke: Good point!  I'll fix that.

melanieclarke:
> A quick glance makes me think that test is checking that the step should raise an exception for an unexpected value. Is that still passing (and raising an exception) because an unknown parameter on the command line is expected to produce an error whereas a parameter in a file should now produce a warning?

Yes, the current change is specifically for parsing configuration files - directly passed parameters on the command line or in step arguments will still raise a ValidationError.  I think this is as requested, to support parameter reference files in CRDS that are a little ahead of the pipeline code.

melanieclarke:
@nden - I don't seem to have permissions on stpipe to request specific reviewers, but I think this one is ready for review, as discussed.

braingram:
The code changes look good to me. I left a comment on [JP-3563](https://jira.stsci.edu/browse/JP-3563) about my concerns for turning this error into a warning.

@melanieclarke would you run the jwst and romancal regtests with this PR?

melanieclarke:
@braingram - I can run the jwst regtests, but I'm not sure if I have access to the romancal ones.

braingram:
Thanks! I started a romancal run here:
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/807/

melanieclarke:
There were a few failures in the jwst regtests, but they're unrelated to this PR.

melanieclarke:
Setting this back to draft to explore warning for extra parameters only when pulling from CRDS.

melanieclarke:
Added some draft changes that would warn for extra values configuration files retrieved from CRDS only.

Note that this adds a validation step earlier in the process than it currently happens, so will add default values to the configuration where they did not exist before.  We'd need to check that this doesn't have downstream effects (i.e. default values overwriting real ones).

Also, I have not added any optional controls for this process yet: it always strips extra values from CRDS when retrieved via get_config_from_reference.  It's not clear to me at what point in the process that might need to be turned off.

In it's current form, there's one downstream unit test in jwst that will need to be updated.

I'm leaving this PR at draft status: it needs more discussion.

melanieclarke:
Closing: the associated JP ticket is withdrawn.

PR #154, opened 2024-06-06
Remove unused code to simplify filename generation

braingram:
This PR removes some unused code in stpipe. Most of this is related to filename generation which will hopefully be somewhat easier to understand after this PR:
- `Pipeline.set_input_filename`
- `Step.name_format`: ~Requires https://github.com/spacetelescope/jwst/pull/8539 to first remove the 1 usage in jwst.~
- `Step.resolve_file_name`
- the `format` argument to `Step.save_model`
- the `name_format` `component_format` and `separator` arguments to `Step._make_output_path`
- remove ([now unused](https://github.com/spacetelescope/jwst/pull/8542)) `Step.reference_uri_to_cache_path`

jwst regtests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1519/ shows 1 unrelated error (`test_duplicate_names` which randomly fails)
romancal regtests: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/825/ passed with no errors

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,10 @@
 ==================
 
 - Remove unused ``Step.closeout`` [#152]
+- Remove unused ``Pipeline.set_input_filename``, ``Step.name_format``,
+  ``Step.resolve_file_name``, ``format`` argument to ``Step.save_model``,
+  ``name_format``, ``component_format`` and ``separator`` arguments to
+  ``Step._make_output_path`` and ``Step.reference_uri_to_cache_path``. [#154]
 
 0.5.2 (2024-03-21)
 ==================
--- a/src/stpipe/pipeline.py
+++ b/src/stpipe/pipeline.py
@@ -245,11 +245,6 @@ class Pipeline(Step):
         config_parser.merge_config(refcfg, pipeline_cfg)
         return refcfg
 
-    def set_input_filename(self, path):
-        self._input_filename = path
-        for key in self.step_defs:
-            getattr(self, key).set_input_filename(path)
-
     def _precache_references(self, input_file):
         """
         Precache all of the expected reference files before the Step's
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -60,12 +60,6 @@ class Step:
     # name.  Must be globally unique!
     class_alias = None
 
-    # String defining the format of the output name, which defines how
-    # **components are inserted into the output file name. If None, use the
-    # default formatting, which is to append Step.suffix to the name. If False,
-    # use basepath as its own format string, containing {suffix}.
-    name_format = None
-
     # Correction parameters. These store and use whatever information a Step
     # may need to perform its operations without re-calculating, or to use
     # from a previous run of the Step.  The structure is up to each Step.
@@ -550,12 +544,10 @@ class Step:
                         if len(results_to_save) <= 1:
                             idx = None
                         if isinstance(result, AbstractDataModel):
-                            self.save_model(result, idx=idx, format=self.name_format)
+                            self.save_model(result, idx=idx)
                         elif hasattr(result, "save"):
                             try:
-                                output_path = self.make_output_path(
-                                    idx=idx, name_format=self.name_format
-                                )
+                                output_path = self.make_output_path(idx=idx)
                             except AttributeError:
                                 self.log.warning(
                                     "`save_results` has been requested, but cannot"
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -623,13 +623,6 @@ class Step:
         """
         raise NotImplementedError("Steps have to override process().")
 
-    def resolve_file_name(self, file_name):
-        """
-        Resolve a file name expressed relative to this Step's
-        configuration file.
-        """
-        return join(dirname(self.config_file or ""), file_name)
-
     @classmethod
     def call(cls, *args, **kwargs):
         """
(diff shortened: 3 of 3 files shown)


braingram:
Closing and re-opening to re-trigger CI (it failed due to github actions being "degraded").

hbushouse:
I can see the benefit in cleaning up a library for unused/unnecessary code, but another part of me thinks it must've had a reason to be there in the first place. Perhaps some of this could still be useful at some point in the future, but just isn't being used right now? Would be nice if we still had access to the thoughts of the original designer and author.

braingram:
Thanks for taking a look.

tldr; this aims to make output file name generation easier to understand

I'll try to fill in details and ping authors (if they're still available).

> Pipeline.set_input_filename

This method was introduced before stpipe existed for the "inital commit" of jwst 8 years ago by @jhunkeler :
https://github.com/spacetelescope/jwst/blame/c353186397091d1821ca8723da49b182346d5626/jwst/stpipe/pipeline.py#L251
It is related to `Step._input_filename` which is set by [Step.set_primary_input](https://github.com/spacetelescope/stpipe/blob/d1a0a6d25de45400e315afa4fa3cee8633a6b545/src/stpipe/step.py#L922)
Under some circumstances `_input_filename` can impact the [default_output_file](https://github.com/spacetelescope/stpipe/blob/d1a0a6d25de45400e315afa4fa3cee8633a6b545/src/stpipe/step.py#L700) which (again under some circumstances) can be used to construct the output path in [_make_output_path](https://github.com/spacetelescope/stpipe/blob/d1a0a6d25de45400e315afa4fa3cee8633a6b545/src/stpipe/step.py#L1104). This is a small portion of the convoluted set of conditions and variables that impact the output filename. Part of me hopes that someone has a better understand of all of this than I do and can help me understand all the conditions that determine how files. I believe removing some of this complication can help to make the file naming easier to understand (and hopefully less buggy). Since `Pipeline.set_input_filename` is unused in stpipe, jwst and romancal removing it is a small step towards this goal.

> Step.name_format

This was introduced by @tapastro in https://github.com/spacetelescope/stpipe/pull/29 to allow for https://github.com/spacetelescope/jwst/pull/6376
It has since been unused by jwst or romancal.
https://github.com/spacetelescope/jwst/pull/8539 used `make_output_path` to accomplish the same goal and I believe can replace any future need for `name_format`.

> Step.resolve_file_name

This was also added in jwst as part of the "initial commit" 8 years ago by @jhunkeler and now appears unused. It includes a single line of code which if needed could be re-introduced (and perhaps updated to use newer syntax):
https://github.com/spacetelescope/stpipe/blob/54ec1d100dd2cd94eddf6848da85e32d3de80aae/src/stpipe/step.py#L639

> the format argument to Step.save_model
   the name_format component_format and separator arguments to Step._make_output_path

These trace back to https://github.com/spacetelescope/jwst/pull/1316 made 6 years ago by @stscieisenhamer The goal in removing these is similar to `Pipeline.set_input_filename`. As these are unknown options that complicate file naming removing them will hopefully make this process easier to understand.

> remove (https://github.com/spacetelescope/jwst/pull/8542) Step.reference_uri_to_cache_path

This change isn't related to file naming. `Step.reference_uri_to_cache_path` is a class method (that doesn't use the class) and contains only a call to:
https://github.com/spacetelescope/stpipe/blob/54ec1d100dd2cd94eddf6848da85e32d3de80aae/src/stpipe/step.py#L921
Any further need for this feature can use the `crds_client` so I don't think the `Step` benefits from having this as a class method.

zacharyburnett (approved):
LGTM, I dunno what `name_format` was used for but if it's not needed that's one less variable-

hbushouse:
@braingram Do any of the changes here result in API changes, such that we need to keep an updated version of stpipe in sync with other packages (e.g. jwst)?

zacharyburnett:
I can't find any usages of `..., name_format=` in the JWST code, but this does change the API so I think it does warrant a version bump

braingram:
Everything removed here is unused in jwst and roman.

This package is 0.5.2 currently so 0.6 seems reasonable to me for the next version (and updating the pins in jwst and roman once this is released to use the new version so `>=0.6`).

I mostly assumed that since this package is not 1.0 that we don't claim any form of backwards compatibility and that as long as we don't break jwst and roman we're good to go. Let me know if this is not the case.

hbushouse:
> Everything removed here is unused in jwst and roman.
> 
> This package is 0.5.2 currently so 0.6 seems reasonable to me for the next version (and updating the pins in jwst and roman once this is released to use the new version so `>=0.6`).
> 
> I mostly assumed that since this package is not 1.0 that we don't claim any form of backwards compatibility and that as long as we don't break jwst and roman we're good to go. Let me know if this is not the case.

Sounds good. Although given the fact that this has been in major public use (by the jwst pipeline) for several years now, I'm wondering why we've never come out with a version 1.0 yet.

braingram:
I think a 1.0 is a great idea but there are some more things I'd like to take a stab at cleaning up before we go there. I'll open an issue for 1.0.0 planning but one major gap is the lack of documentation.

hbushouse:
> I think a 1.0 is a great idea but there are some more things I'd like to take a stab at cleaning up before we go there. I'll open an issue for 1.0.0 planning but one major gap is the lack of documentation.

Good point. So for the impending release we'll stick with 0.6.0, but start making plans for a future 1.0.0.

PR #163, opened 2024-06-26
remove Windows tests from CI

### LINKED ISSUE #162: Use of `os.uname` causes failures on windows
`os.uname` [does not exist on windows](https://docs.python.org/3/library/os.html#os.uname). It's usage in this package:
https://github.com/spacetelescope/stpipe/blob/915a8326432254bd5d03ee67586bc94c79d74327/src/stpipe/cmdline.py#L350-L351
causes failures on windows:
https://github.com/spacetelescope/stpipe/actions/runs/9637434273/job/26576657996#step:10:400

Some clarification on if we support windows (romancal and jwst do not) would be helpful. If we do support windows we'll need a replacement for `os.uname`. If we do not support windows we could disable the windows tests.

zacharyburnett:
closes #162 

since `jwst` and `romancal` and `hstcal` do not support Windows, we might as well stop testing on Windows here

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,6 @@
 [![codecov](https://codecov.io/gh/spacetelescope/stpipe/branch/main/graph/badge.svg?token=Mm9I0X1o4X)](https://codecov.io/gh/spacetelescope/stpipe)
 
 Provides base classes and command-line tools for implementing calibration pipeline software.
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


braingram:
Thanks! Would you also add a note to the readme similar to jwst and romancal:
```
> [!NOTE]
> Linux and MacOS platforms are tested and supported.  Windows is not currently supported.
```
and a changelog entry?

PR #168, opened 2024-07-22
`AbstractModelLibrary` docs

braingram:
This PR adds additional user and developer documentation for `ModelLibrary`.

Several other docs are updated as this PR also set's a [default_role](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role) of 'obj' for sphinx. This turns all default backtick annotated text into `py:obj` roles. This was done to make writing the `ModelLibrary` docs easier but also fixed quite a few links in the other docs. This does mean that literal text needs 2 backticks (as in seen in several of the non-library doc changes in this PR).

Link to generated docs: https://stpipe--168.org.readthedocs.build/en/168/

### DIFF
--- a/docs/source/model_library.rst
+++ b/docs/source/model_library.rst
@@ -161,7 +161,7 @@ Although the specifics of what is returned by `~stpipe.library.AbstractModelLibr
 depends on how the subclass implements ``AbstractModelLibrary._load_asn``, it
 is required that the association metadata dictionary contain a "members" list. This
 can be inspected via ``library.asn["products"][0]["members"]`` and must contain a
-dictionary for each "member" including key value pairs for:
+dictionary for each "member" including key-value pairs for:
 
 - "expname" for the exposure name, with a string value corresponding to the
   name of the file for this member
--- a/docs/source/model_library.rst
+++ b/docs/source/model_library.rst
@@ -403,7 +403,8 @@ A survey of container usage was performed with a few key findings:
 - When loaded from an association, the container size never changed;
   that is, no use-cases required adding new models to associations within steps
 - The order of models was never changed
-- Needs various methods for stpipe
+- Must be compatible with stpipe infrastructure (eg implements
+  ``crds_observatory``, ``get_crds_parameters``, etc methods)
 - Several steps implemented different memory optimizations
   and had significant complexity added to deal with containers
   that sometimes returned filenames and sometimes returned models
--- a/docs/source/model_library.rst
+++ b/docs/source/model_library.rst
@@ -90,6 +90,15 @@ library that the model was not modified.
 This tells the library not to overwrite the model's temporary file while shelving, saving
 on both disk space and the time required to write.
 
+.. WARNING::
+   In the above example ``model`` remains in scope after the call to
+   `~stpipe.library.AbstractModelLibrary.shelve` (and even after
+   the exit of the with statement). This means ``model`` will not
+   be garbage collected (and it's memory will not be freed) until
+   the end of the scope containing the ``with library`` exits. If
+   more work occurs within the scope please consider adding an
+   explicit ``del model`` when your code is finished with the model.
+
 
 .. _library_map_function:
 
--- a/docs/source/model_library.rst
+++ b/docs/source/model_library.rst
@@ -168,7 +168,7 @@ dictionary for each "member" including key value pairs for:
 - "exptype" for the exposure type with a string value describing the type
   of exposure (for example "science" or "background")
 
-Although not required "group_id" (with a string value corresponding to the
+Although not required, "group_id" (with a string value corresponding to the
 group name) should be added to each member dictionary (see
 :ref:`library_association` for more details).
 
--- a/docs/source/model_library.rst
+++ b/docs/source/model_library.rst
@@ -3,14 +3,14 @@
 Model Library
 =============
 
-`AbstractModelLibrary` is a container designed to allow efficient processing of
-collections of `AbstractDataModel` instances created from an association.
+`~stpipe.library.AbstractModelLibrary` is a container designed to allow efficient processing of
+collections of `~stpipe.datamodel.AbstractDataModel` instances created from an association.
 
-`AbstractModelLibrary` is an ordered collection (like a `list`) but provides:
+`~stpipe.library.AbstractModelLibrary` is an ordered collection (like a `list`) but provides:
 
-- access to association metadata: `AbstractModelLibrary.asn`
-- grouping API: `AbstractModelLibrary.group_indices` and `AbstractModelLibrary.group_names`
-- compatibility with `Step` and `Pipeline` runs
+- access to association metadata: `~stpipe.library.AbstractModelLibrary.asn`
+- grouping API: `~stpipe.library.AbstractModelLibrary.group_indices` and `~stpipe.library.AbstractModelLibrary.group_names`
+- compatibility with `~stpipe.step.Step` and `~stpipe.pipeline.Pipeline` runs
 - a consistent indexing API that is the same for "in memory" and "on disk" libraries
 
 
@@ -34,7 +34,7 @@ Most commonly an instance will be created from an association file:
 Borrowing and shelving models
 -----------------------------
 
-Interacting with an `AbstractModelLibrary` involves "borrowing" and "shelving"
+Interacting with an `~stpipe.library.AbstractModelLibrary` involves "borrowing" and "shelving"
 models, both of which must occur during a ``with`` statement (while the library
 is "open"):
 
@@ -77,7 +77,7 @@ containing the models.
    point will the library overwrite the input file.
 
 If model is not modified during the time it's borrowed (for example if the
-`model.dq` array was read, but not modified). It is helpful to tell the
+``model.dq`` array was read, but not modified). It is helpful to tell the
 library that the model was not modified.
 
 .. code-block:: pycon
@@ -98,7 +98,7 @@ Map function
 
 Let's say you want to get the ``meta.filename`` attribute for all models
 in a library. The above "open", "borrow", "shelve" pattern can be quite
-verbose. Instead, the helper method `AbstractModelLibrary.map_function`
+verbose. Instead, the helper method `~stpipe.library.AbstractModelLibrary.map_function`
 can be used to generate an iterator that returns the result of a function
 applied to each model in the library:
 
@@ -110,7 +110,7 @@ applied to each model in the library:
    >>> filenames = list(library.map_function(get_model_name))
 
 .. NOTE::
-   `AbstractModelLibrary.map_function` does not require an open library
+   `~stpipe.library.AbstractModelLibrary.map_function` does not require an open library
    and will handle opening, borrowing, shelving and closing for you.
 
 
@@ -130,12 +130,12 @@ performed on the association metadata).
    ...     print(f"\tModel indices for {group_name}: {group_index_map[group_name]}")
 
 .. WARNING::
-   Although `AbstractModelLibrary.group_names` and
-   `AbstractModelLibrary.group_indices` do not require an open library,
+   Although `~stpipe.library.AbstractModelLibrary.group_names` and
+   `~stpipe.library.AbstractModelLibrary.group_indices` do not require an open library,
    any "borrows" using the indices do. Be sure to open the library before
    trying to borrow a model.
 
-`AbstractModelLibrary.asn` provides read-only access to the association data.
+`~stpipe.library.AbstractModelLibrary.asn` provides read-only access to the association data.
 
 .. code-block:: pycon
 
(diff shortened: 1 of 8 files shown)


docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+
+   >>> library = ModelLibrary("my_big_asn.json", on_disk=True)
+   >>> with library:
+   ...     model = library.borrow(0)  # the input file for model 0 is loaded
+   ...     library.shelve(model)  # a temporary file for model 0 is written
+
+.. NOTE::
+   In the above example, a temporary file was created for model 0. At no
+   point will the library overwrite the input file.
+
```
emolter, line 78: it might be good to note somewhere here that `model` is still in memory, e.g.,
```
with library:
    model = library.borrow(0)
    library.shelve(model, 0)
#do something with model
```
works.  This was unintuitive to me at first, the fact that the shelve command is NOT equivalent to `del model` in a case like this.

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+   >>> def get_model_name(model, index):
+   ...     return model.meta.filename
+   >>>
+   >>> filenames = list(library.map_function(get_model_name))
+
+.. NOTE::
+   `~stpipe.library.AbstractModelLibrary.map_function` does not require an open library
+   and will handle opening, borrowing, shelving and closing for you.
+
```
emolter, line 115: might be good to specify here that the arguments to map_function MUST be `model, index`.  And random question too, does it need to take exactly those arguments, or can you have any number of arguments as long as those are the first two?

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+
+.. _library_grouping:
+
+Grouping
```
emolter, line 119: it occurs to me at this point that it might be good to have an additional section (perhaps very near the top) that just discusses how to access the association metadata, i.e., what the names of all the metadata attributes are and what they return.

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+   referenced in the association be modified. However, the above pattern
+   does allow ``Step.process`` to "modify" ``init`` when
+   ``init`` is a `~stpipe.library.AbstractModelLibrary` (the models
+   in the library will not be copied).
+
+``Step.process`` can extend the above pattern to
+support additional inputs (for example a single
+`~stpipe.datamodel.AbstractDataModel` or filename containing
+a `~stpipe.datamodel.AbstractDataModel`) to allow more
+flexible data processings. Although some consideration
```
emolter, line 209:
```suggestion
flexible data processings, although some consideration
```

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+
+Isolated Processing
+-------------------
+
+Let's say we have a `~stpipe.step.Step`, ``flux_calibration``
+that performs an operation that is only concerned with the data
+for a single `~stpipe.datamodel.AbstractDataModel` at a time.
+This step applies a function ``calibrate_model_flux`` that
+accepts a single `~stpipe.datamodel.AbstractDataModel` and index as an input.
+It's ``Step.process`` function can make good use of
```
emolter, line 228:
```suggestion
Its ``Step.process`` function can make good use of
```

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+
+   ClosedLibraryError: ModelLibrary is not open
+
+The library must be "open" (used in a ``with`` statement) before
+a model can be borrowed. This is important for keeping track of
+which models were possibly modified.
+
+This error can be avoided by "opening" the library before calling
+`~stpipe.library.AbstractModelLibrary.borrow` (and being sure to call
+`~stpipe.library.AbstractModelLibrary.shelve`, more on that below):
```
emolter, line 264:
```suggestion
`~stpipe.library.AbstractModelLibrary.shelve`):
```
my logic is that it's unlikely someone will be reading this section linearly - it's likely to be a reference if someone runs into these errors.

Random note, but I'm impressed by how helpful the various error messages have been during my own development with ModelLibrary - nice work!

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+
+.. _library_developer_documentation:
+
+Developer Documentation
+=======================
+
+What follows are note primarily aimed towards developers and
+maintainers of `~stpipe.library.AbstractModelLibrary`. This section might be useful
+to provide context to users but shouldn't be necessary for a user
+to effectively use `~stpipe.library.AbstractModelLibrary`.
```
emolter, line 382:
```suggestion
```
Not sure any of this is necessary, it seems implied by the header

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+to be overwritten. It was decided that a new container would be
+developed to address these and other issues. This would allow
+gradual migration for pipeline code where specific steps and pipelines
+could update to `~stpipe.library.AbstractModelLibrary` while leaving the existing
+container unchanged for other steps.
+
+A survey of container usage was performed with a few key findings:
+
+- Many uses could be replaced by simpler containers (lists)
+- When loaded from an association, the container size never changed
```
emolter, line 403:
```suggestion
- When loaded from an association, the container size never changed;
  that is, no use-cases required adding new models to associations within steps
```

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+- The order of models was never changed
+- Needs various methods for stpipe
```
emolter, line 405: what does this line mean?

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+- Several steps implemented different memory optimizations
```
emolter, line 406: this feels like two bullet points

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+  container to pass between steps.
+
+Most of the core functionality is public and described in the above
+user documentation. What follows will be description of other parts
+of the API (most private) and internal details.
+
+One core issue is how can the container know when to load and
+save models (to temporary files) if needed? With a typical list
+``__getitem__`` can map to load but what will map to save?
+Initial prototypes used ``__setitem__`` which lead to some confusion
```
emolter, line 440:
```suggestion
Initial prototypes used ``__setitem__`` which led to some confusion
```

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+A similar note of caution applies here where some consideration
+of the complexity required vs the benefits is recommended.
+
+.. _library_get_sections:
+
+Get sections
+^^^^^^^^^^^^
+
+`~stpipe.library.AbstractModelLibrary` has no replacement for
+the ``get_sections`` API provided with ``ModelContainer``. If it's use
```
emolter, line 534:
```suggestion
the ``get_sections`` API provided with ``ModelContainer``. If its use
```

emolter (commented):
Thanks for all your hard work on this, Brett, it's really useful!

docs/source/model_library.rst
```diff
@@ -0,0 +1,550 @@
...
+   ...    model = library.borrow(0)
+   ...    # do stuff with the model...
+   ...    library.shelve(model)
+
+Iteration is also supported (but don't forget to return your models!).
+
+.. code-block:: pycon
+
+   >>> with library:
+   ...    for model in library:
```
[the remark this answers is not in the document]
braingram: Thanks! Added in a2a53bd (with a minor formatting change)

docs/source/model_library.rst
braingram: I changed this bullet in 79517d3

Is the new wording more clear?

docs/source/model_library.rst
braingram: Good point. I added this as one because the complexity was due to the memory optimization (if the container always returned either a filename or a model the step code would be simpler). Does the split in f350f55 work?

docs/source/model_library.rst
emolter: sure, that's fine with me

docs/source/model_library.rst
braingram: Good point! How does the warning added here look to you?: b484e29

docs/source/model_library.rst
emolter: yeah i think this is fine

docs/source/model_library.rst
emolter: yes, thanks, that's much clearer. just one grammar thing is that e.g. should have dots

docs/source/model_library.rst
braingram: Is the docstring for the function sufficient? If not I'd rather expand on that (since it's what will be seen in an interactive session).

At the moment the only supported function signature is `model, index`. Would you open a separate issue to discuss changes to this? I think it's worth considering changing this as we get more experience with `ModelLibrary` including:
- making `index` optional
- adding optional `args` and `kwargs` that get passed through `map_function`

docs/source/model_library.rst
braingram: Thanks. I removed the eg (since it already has etc) in 8a01ef2

docs/source/model_library.rst
emolter: yes, can do.  I haven't played with `map_function` myself yet, but I will try to replace some of my explicit `borrow` and `shelve` loops with it in the JWST PR.  Then it will be easier to point to specific examples where this functionality might be useful

docs/source/model_library.rst
braingram: I expanded this a bit in 55ccf14

I don't think we can describe all the association metadata here (as there are differences between roman and jwst). I attempted to describe the common structure required by `ModelLibrary`.

docs/source/model_library.rst
```diff
@@ -144,13 +144,34 @@ performed on the association metadata).
...
+is required that the association metadata dictionary contain a "members" list. This
+can be inspected via ``library.asn["products"][0]["members"]`` and must contain a
+dictionary for each "member" including key value pairs for:
+
+- "expname" for the exposure name with a string value corresponding to the
+  name of the file for this member
+- "exptype" for the exposure type with a string value describing the type
+  of exposure (for example "science" or "background")
+
+Although not required "group_id" (with a string value corresponding to the
```
emolter, line 171:
```suggestion
Although not required, "group_id" (with a string value corresponding to the
```

docs/source/model_library.rst
```diff
@@ -144,13 +144,34 @@ performed on the association metadata).
...
    >>> library.asn["products"][0]["name"]
    >>> library.asn["table_name"]
 
+Although the specifics of what is returned by `~stpipe.library.AbstractModelLibrary.asn`
+depends on how the subclass implements ``AbstractModelLibrary._load_asn``, it
+is required that the association metadata dictionary contain a "members" list. This
+can be inspected via ``library.asn["products"][0]["members"]`` and must contain a
+dictionary for each "member" including key value pairs for:
+
+- "expname" for the exposure name with a string value corresponding to the
```
emolter, line 166:
```suggestion
- "expname" for the exposure name, with a string value corresponding to the
```

docs/source/model_library.rst
```diff
@@ -144,13 +144,34 @@ performed on the association metadata).
...
+  name of the file for this member
+- "exptype" for the exposure type with a string value describing the type
```
emolter, line 168:
```suggestion
- "exptype" for the exposure type, with a string value describing the type
```

docs/source/model_library.rst
```diff
@@ -144,13 +144,34 @@ performed on the association metadata).
...
 .. code-block:: pycon
 
    >>> library.asn["products"][0]["name"]
    >>> library.asn["table_name"]
 
+Although the specifics of what is returned by `~stpipe.library.AbstractModelLibrary.asn`
+depends on how the subclass implements ``AbstractModelLibrary._load_asn``, it
+is required that the association metadata dictionary contain a "members" list. This
+can be inspected via ``library.asn["products"][0]["members"]`` and must contain a
+dictionary for each "member" including key value pairs for:
```
emolter, line 164:
```suggestion
dictionary for each "member" including key-value pairs for:
```

docs/source/model_library.rst
```diff
@@ -144,13 +144,34 @@ performed on the association metadata).
...
 
 .. code-block:: pycon
 
    >>> library.asn["products"][0]["name"]
    >>> library.asn["table_name"]
 
+Although the specifics of what is returned by `~stpipe.library.AbstractModelLibrary.asn`
+depends on how the subclass implements ``AbstractModelLibrary._load_asn``, it
+is required that the association metadata dictionary contain a "members" list. This
+can be inspected via ``library.asn["products"][0]["members"]`` and must contain a
```
emolter, line 163: what does the index `[0]` correspond to?

docs/source/model_library.rst
emolter: yeah I figured that would be the case.  What you did is already super helpful.  I can try to fill in the gaps when documenting the JWST side

docs/source/model_library.rst
braingram: The 0th index (only) product since "products" is a list: https://jwst-pipeline.readthedocs.io/en/latest/jwst/associations/level3_asn_technical.html#example-association

emolter (approved):
All my questions and comments have been addressed

braingram:
Thanks @emolter for all your help!

PR #171, opened 2024-07-23
conditionally format log_records

braingram:
Instead of storing every log message as a record,
only store the formatted logs (if a formatter was provided). A formatter can be supplied by setting `_log_records_formatter` on the `Step` class/instance.


src/stpipe/log.py
```diff
@@ -280,18 +280,24 @@ def log_records(self):
         return self._log_records
 
     def emit(self, record):
-        self._log_records.append(record)
+        if not self.formatter:
+            return
+        self._log_records.append(self.formatter.format(record))
```
zacharyburnett, line 285: can we make this more explicit?
```suggestion
        if self.formatter is not None:
            self._log_records.append(self.formatter.format(record))
```

zacharyburnett (approved):
LGTM, had a comment

src/stpipe/log.py
braingram: Thanks!

jdavies-st:
This PR is good.  Perhaps for a future PR, I would go one step further and get rid of any `Step` attributes that point to the log.  There's no reason to store the log as an attribute in a class, as logs are global singletons.  This is an anti-pattern to avoid:

https://docs.python.org/3/howto/logging-cookbook.html#using-loggers-as-attributes-in-a-class-or-passing-them-as-parameters

So really, there should not be a `self.log` in `Step`, nor should there be `self.log_records` at all.

PR #174, opened 2024-07-25
Log records patch

braingram:
Cherry pick changes from #171 to 0.6.0 release branch.

I made the 0.6.0 release branch for this PR so please also check if the release branch looks ok (I made it from the `0.6.0` tag).

After this is merged, the release plan (as discussed with @zacharyburnett) is:
- make patch release of romancal with stpipe pin <0.6.1: https://github.com/spacetelescope/romancal/pull/1331
- make stpipe 0.6.1 release from `release/0.6.1`

PR #175, opened 2024-07-30
Revert "Log records patch"

braingram:
Reverts spacetelescope/stpipe#174


zacharyburnett:
we could also rebase the `release/0.6.x` branch to drop that commit, to keep the commit history clean

braingram:
> we could also rebase the `release/0.6.x` branch to drop that commit, to keep the commit history clean

Good idea! I was lazy and clicked the 'revert' button on github :)

zacharyburnett (approved):

zacharyburnett:
ok, commit dropped

Issue #173, opened 2024-07-24
Explicitly `del` model in `ModelLibrary.map_function`


braingram:
In the following loop:
https://github.com/spacetelescope/stpipe/blob/9952c365458fc3f01839a912b62ab6c637dc9179/src/stpipe/library.py#L739-L745
The lack of an explicit `del model` after the shelve keeps it in scope until it's overwritten by the `enumerate` iterator which only overwrites it after the next model is loaded. This results in 2 models in memory when only 1 is needed.

PR #181, opened 2024-08-07
del model in map_function

braingram:
This PR makes a minor change to `AbstractModelLibrary.map_function` to reduce memory usage. More details can be found in the issue but the tldr is that this prevents having 2x models in memory during `map_function`.

Fixes: https://github.com/spacetelescope/stpipe/issues/173


zacharyburnett (approved):

braingram:
Thanks! The romancal failure is due to the unrelated background schema changes.

PR #190, opened 2024-09-26
JP-3721: Simplify ModelContainer

emolter:
Helps to resolve [JP-3721](https://jira.stsci.edu/browse/JP-3721)

Helps to close https://github.com/spacetelescope/jwst/issues/8738 

This PR enables the JWST ModelContainer to no longer inherit from DataModel and to no longer require its own `save` method, as part of an effort to make ModelContainer's usage narrower and easier to understand.  This required two changes to the logic of step.py:

- Add handling for Sequence when setting `skip` attribute
- In save_model(), remove the assumption that a Sequence-type object will always have a `save` method

## Tasks

</details>

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -492,9 +492,11 @@ class Step:
                                     )
                                 library.shelve(model, i)
 
-                    elif (isinstance(args[0], Sequence)) and \
-                         (not isinstance(args[0], str)) and \
-                         (self.class_alias is not None):
+                    elif (
+                        (isinstance(args[0], Sequence))
+                        and (not isinstance(args[0], str))
+                        and (self.class_alias is not None)
+                    ):
                         # handle ModelContainer or list of models
                         if args[0] and isinstance(args[0][0], AbstractDataModel):
                             for model in args[0]:
@@ -509,12 +511,12 @@ class Step:
                                         e,
                                     )
 
-                    elif isinstance(args[0], AbstractDataModel) and \
-                        self.class_alias is not None:
+                    elif (
+                        isinstance(args[0], AbstractDataModel)
+                        and self.class_alias is not None
+                    ):
                         try:
-                            args[0][
-                                f"meta.cal_step.{self.class_alias}"
-                            ] = "SKIPPED"
+                            args[0][f"meta.cal_step.{self.class_alias}"] = "SKIPPED"
                         except AttributeError as e:
                             self.log.info(
                                 "Could not record skip into DataModel header: %s",
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -560,34 +560,7 @@ class Step:
 
                 # Save the output file if one was specified
                 if not self.skip and self.save_results:
-                    # Setup the save list.
-                    if not isinstance(step_result, Sequence):
-                        results_to_save = [step_result]
-                    else:
-                        results_to_save = step_result
-
-                    for idx, result in enumerate(results_to_save):
-                        if len(results_to_save) <= 1:
-                            idx = None
-                        if isinstance(
-                            result, (AbstractDataModel | AbstractModelLibrary)
-                        ):
-                            self.save_model(result, idx=idx)
-                        elif hasattr(result, "save"):
-                            try:
-                                output_path = self.make_output_path(idx=idx)
-                            except AttributeError:
-                                self.log.warning(
-                                    "`save_results` has been requested, but cannot"
-                                    " determine filename."
-                                )
-                                self.log.warning(
-                                    "Specify an output file with `--output_file` or set"
-                                    " `--save_results=false`"
-                                )
-                            else:
-                                self.log.info("Saving file %s", output_path)
-                                result.save(output_path, overwrite=True)
+                    self.save_model(step_result)
 
                 if not self.skip:
                     self.log.info("Step %s done", self.name)
@@ -995,6 +968,9 @@ class Step:
         if not force and not self.save_results and not output_file:
             return None
 
+        if model is None:
+            return None
+
         if isinstance(model, AbstractModelLibrary):
             output_paths = []
             with model:
@@ -1013,20 +989,36 @@ class Step:
             return output_paths
 
         elif isinstance(model, Sequence):
-            output_paths = []
-            for i, m in enumerate(model):
-                output_paths.append(
-                    self.save_model(
-                        m,
-                        idx=i,
-                        suffix=suffix,
-                        force=force,
-                        **components,
+            if not hasattr(model, "save"):
+                # list of datamodels, e.g. ModelContainer
+                output_paths = []
+                for i, m in enumerate(model):
+                    idx = None if len(model) == 1 else i
+                    output_paths.append(
+                        self.save_model(
+                            m,
+                            idx=idx,
+                            suffix=suffix,
+                            force=force,
+                            **components,
+                        )
                     )
+                return output_paths
+            else:
+                # JWST SourceModelContainer takes this path
+                save_model_func = partial(
+                    self.save_model,
+                    suffix=suffix,
+                    force=force,
+                    **components,
                 )
-            return output_paths
+                output_path = model.save(
+                    path=output_file,
+                    save_model_func=save_model_func,
+                )
+                return output_path
 
-        else:
+        elif hasattr(model, "save"):
             # Search for an output file name.
             if self.output_use_model or (
                 output_file is None and not self.search_output_file
@@ -1042,8 +1034,10 @@ class Step:
                 )
             )
             self.log.info("Saved model in %s", output_path)
+            return output_path
 
-        return output_path
+        else:
+            return
 
     @property
     def make_output_path(self):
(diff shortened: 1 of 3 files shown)


emolter:
starting regression tests on jwst main with this PR branch pinned as the stpipe version.  This should show whether the stpipe changes are back-compatible with the previous ModelContainer

emolter:
Only one failed regtest, but starting a full new run just in case.

There are currently very few tests of `save_model()` and I'm looking at adding some more.

emolter:
I have now added several unit tests of the step `save()` method.  Hopefully these, along with the corresponding changes in jwst https://github.com/spacetelescope/jwst/pull/8831, clarify the intent behind this PR.

The passing regression tests in the link above, which use the jwst main branch with stpipe pinned to this PR branch show that these changes are compatible with that pipeline even without making any changes to ModelContainer.  Therefore I think we could safely merge this without waiting for the corresponding changes in the jwst repository.  I don't think the two failing unit tests in jwst test_downstream are caused by this PR, because the same ones are also present on [yesterday's CI on main](https://github.com/spacetelescope/stpipe/actions/runs/11212199737/job/31162725683).

Marking this ready for review now.

Can someone please start a romancal regtest run with this branch at some point?  It would be good to ensure we aren't breaking anything there, either.

src/stpipe/step.py
```diff
@@ -507,16 +507,15 @@ def run(self, *args):
...
                         try:
-                            args[0][
-                                f"meta.cal_step.{self.class_alias}"
-                            ] = "SKIPPED"
+                            args[0][f"meta.cal_step.{self.class_alias}"] = "SKIPPED"
                         except AttributeError as e:
                             self.log.info(
-                                "Could not record skip into DataModel"
-                                " header: %s",
+                                "Could not record skip into DataModel" " header: %s",
```
braingram, line 518:
```suggestion
                                "Could not record skip into DataModel header: %s",
```

braingram (commented):
Here are some minor comments (that I found I had not previously submitted).

I haven't had a chance to look at this PR in detail but I'm surprised by the number of required changes to simplify ModelContainer. Is the expectation that stpipe should be able to treat it like a list of models?

emolter:
> I haven't had a chance to look at this PR in detail but I'm surprised by the number of required changes to simplify ModelContainer. Is the expectation that stpipe should be able to treat it like a list of models?

Yes, in the present version of the code a list of datamodels and the new ModelContainer take the same logic path.  Some more notes on what changed:

- The refactor removes repeated or near-repeated code between `step.run()` and `step.save_model()` that were both designed to handle lists of models.  I don't know if `ModelContainer` was the original reason this was all implemented, but it doesn't seem necessary to me and handling everything inside `save_model` made more sense in my head.
- It was necessary to also allow a Sequence that had its own `save()` attribute because `SourceModelContainer` has one.  I can see the argument for removing this too, but the problem is that `SourceModelContainer` encodes a `MultiExposureModel` and therefore is in some ways akin to a singular datamodel.  There is an open [issue](https://github.com/spacetelescope/jwst/issues/8742) about rethinking `SourceModelContainer` and I'd love to hear your thoughts, but the more I think about it, the more I think it's mostly fine as-is.
- I decided to make it impossible to save lists-of-lists.  The motivation for that was the jwst badpix_selfcal step, for which allowing Sequence-type input in the way I'm doing it meant that `step.save_model()` was saving the science exposure and all background exposures to the same filename.  In a previous version, I had modified `save_model` for just the `badpix_selfcal_step`, but that made it so the stpipe changes were not compatible with jwst/main.  This decision could be changed though.

emolter:
> Are you saying the code on main isn't behaving as expected so this PR is also fixing a different bug?

No, but it's behaving in an undefined way.

> Can this be handled in jwst (either in the Step subclass or by fixing SourceModelContainer to have an API consistent with ModelContainer)? That way the code here that's shared with roman can be kept generic.

Not sure what isn't generic about the present solution, as it does not check for any jwst-specific data types, just a `save` method.  It's also required to maintain support for the old version of ModelContainer.

> Is is straightforward to undo that de-duplication? It sounds like it's not required for this PR and would hopefully make it easier to review.

I can try.

braingram:
> No, but it's behaving in an undefined way.

What's undefined? I'm expecting the answer might be the lack of documentation.

> Not sure what isn't generic about the present solution, as it does not check for any jwst-specific data types, just a `save` method. It's also required to maintain support for the old version of ModelContainer.

I sort of expected that a simplification of ModelContainer in jwst wouldn't involve changes to stpipe (or at least wouldn't add code here since ModelContainer isn't used in roman). The comment `# JWST SourceModelContainer takes this path` indicates that there is a logical path being added that will never be used by roman. Even though this isn't checking specifically for a jwst type it (and the other paths that handle only ModelContainer) look designed to only handle the jwst types. Would it be easier to code and read to check for those types? This would be easier if some portion of this step code was implemented in the jwst Step subclass (which might require some refactoring here to provide methods that the subclass can override).
 
I'd be curious to hear your thoughts on either:
- implementing this jwst specific logic in the jwst Step subclass
- retaining the ModelContainer API that stpipe expects. This would put a limit on what simplifcation is possible within jwst. Are all of the changes here a result of dropping `save` from ModelContainer?

emolter:
have a look at https://github.com/spacetelescope/stpipe/pull/194/files
Does this look more like what you expected?

I think I was using the refactor as a crutch to understand the code better.  I think you're right that this was possible with many fewer changes to step.py.

braingram:
Thanks for updating this PR. I think my previous comment and question got missed. My main question is are all of the changes here a result of dropping save from ModelContainer?

emolter:
> are all of the changes here a result of dropping save from ModelContainer?

That's part of it, but mainly it's dropping the inheritance of AbstractDataModel from ModelContainer.  It's true that ModelContainer used to be saved by its own `save` method, but that only worked within stpipe because it was being caught by `isinstance(model, AbstractDataModel)`

braingram:
Thanks. If save is kept does the simplified ModelContainer pass an `isinstance` check against `AbstractDataModel`? I think so and hopefully that would mean no changes are needed here in stpipe. It sounds like SourceModelContainer still needs a save so I'm in favor of keeping save on ModelContainer (especially if it means no changes are needed here).

emolter:
> If save is kept does the simplified ModelContainer pass an isinstance check against AbstractDataModel

Not sure what you're asking, of course it doesn't pass such a check unless it's inheriting from AbstractDataModel, which it is not.  Those checks can (and, basically, have) been changed to a check whether the input has its own `save` method.  I can see an argument for only checking that, and not whether it's a datamodel at all.

I am not against retaining a `save` method for ModelContainer.  However, I don't think that would mean zero changes to stpipe, and to be frank I still do not understand the hesitancy to make the proposed changes.

braingram:
> Not sure what you're asking, of course it doesn't pass such a check unless it's inheriting from AbstractDataModel, which it is not. Those checks can (and, basically, have) been changed to a check whether the input has its own `save` method. I can see an argument for only checking that, and not whether it's a datamodel at all.

The [__subclasshook__](https://github.com/spacetelescope/stpipe/blob/ecd5d162be425c24db2498b34bcbaeccec4ac557/src/stpipe/datamodel.py#L19) in AbstractDataModel means explicit inheritance isn't required. Here's an example:
```python
import stpipe

class NotAModel:
    def save(self):
        pass

    @property
    def crds_observatory(self):
        pass

    def get_crds_parameters(self):
        pass

assert isinstance(NotAModel(), stpipe.datamodel.AbstractDataModel)
```
The assert succeeds even though `NotAModel` doesn't explicitly inherit from `AbstractDataModel`. This is how stpipe can check for a "DataModel" even though neither JwstDataModel or RomanDataModel inherit from `AbstractDataModel`.

> I am not against retaining a `save` method for ModelContainer. However, I don't think that would mean zero changes to stpipe, and to be frank I still do not understand the hesitancy to make the proposed changes.

My hesitation is that stpipe is mostly unmaintained and a critical part of jwst and roman. It's pretty lacking in tests (it might be a bug in the coverage but this PR shows only 50% of the diff covered) and especially for roman the tests in the downstream packages don't well cover all the expected features from stpipe. Given that the motivation for this change is entirely due to jwst it seems overly risky to change stpipe if not necessary.

emolter:
> The [subclasshook](https://github.com/spacetelescope/stpipe/blob/ecd5d162be425c24db2498b34bcbaeccec4ac557/src/stpipe/datamodel.py#L19) in AbstractDataModel means explicit inheritance isn't required.

Well, that would have been very good for me to know at the start of all of this.  Learning hurts I guess...

With that knowledge it does seem likely that retaining `save` would allow stpipe to go largely unchanged.  I will try that out.

> My hesitation is that stpipe is mostly unmaintained and a critical part of jwst and roman. It's pretty lacking in tests (it might be a bug in the coverage but this PR shows only 50% of the diff covered) and especially for roman the tests in the downstream packages don't well cover all the expected features from stpipe

The way I'm hearing this is, "stpipe should be maintained and tests should be expanded to cover any changes."  Historically things [have changed](https://github.com/spacetelescope/stpipe/commit/8550a7482e3f16ac0680f3a31a59d2ed06094924) as a result of ModelContainer specifically.  But maintaining the repository requires a human person, of course, so it's a management question and probably beyond either of our control.

If I can make things work without changes to `step.run()`, I will change this PR to just add the unit tests.  Do you have any comments about those?

PR #192, opened 2024-10-04
JP-3490: Allow run to check CRDS or user parameters for default overrides

melanieclarke:
Resolves [JP-3490](https://jira.stsci.edu/browse/JP-3490)

Closes #

This PR proposes a way to resolve the long-standing issue that calling pipeline or step via `run` retrieves different parameter defaults than calling it via the class method `call`.  The idea is to keep run and call as is, but allow run to check CRDS for parameter overrides when called.

To allow users to set parameters as attributes and not override them with defaults, we can track initialization status for all parameters.
- Override __setattr__ to set initialized = True whenever a parameter is set.
- In __init__, set initialized = False for all parameters set from software defaults in the spec

Inside run, check for CRDS defaults and use them.
- Use the provided input data to retrieve any param files from CRDS, and build a config in the same way `call` does.
- Set the parameters to the CRDS config defaults, for any that have not already been initialized.

This would be mostly invisible to users: both call and run would work in the same way they currently do, except that recommended defaults are retrieved from CRDS when run is called.

Additionally, since we are updating a config inside run, we can also allow it to accept `kwargs` in the same way call does, so users could directly set parameters for steps when calling run. Since run and call can use the same method to retrieve and merge CRDS and user parameters, this is straightforward to support.

If this PR is accepted, the JWST documentation will need to be updated, in [readthedocs](https://jwst-pipeline.readthedocs.io/en/stable/jwst/stpipe) and in [JDox](https://jwst-docs.stsci.edu/jwst-science-calibration-pipeline/running-the-jwst-science-calibration-pipeline#RunningtheJWSTScienceCalibrationPipeline-RunningfromwithinPython), to note the change to behavior for `run`.

The romancal documentation will also need updating, here: 
https://roman-pipeline.readthedocs.io/en/latest/roman/stpipe/call_via_run.html

## Tasks

</details>

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -334,6 +334,7 @@ class Step:
             Additional parameters to set.  These will be set as member
             variables on the new Step instance.
         """
+        self._initialized = {}
         self._reference_files_used = []
         # A list of logging.LogRecord emitted to the stpipe root logger
         # during the most recent call to Step.run.
@@ -361,6 +362,11 @@ class Step:
         # Set the parameters as member variables
         for key, val in kws.items():
             setattr(self, key, val)
+
+            # Mark them as uninitialized, from the user standpoint,
+            # unless they are required to be used as is
+            if _validate_kwds:
+                self._initialized[key] = False
 
         # Create a new logger for this step
         self.log = log.getLogger(self.qualified_name)
@@ -402,6 +408,12 @@ class Step:
                     i,
                 )
 
+    def __setattr__(self, key, value):
+        """Override setattr to track initialization status for step parameters."""
+        super().__setattr__(key, value)
+        if not key.startswith('_'):
+            self._initialized[key] = True
+
     @property
     def log_records(self):
         """
@@ -413,7 +425,7 @@ class Step:
         """
         return self._log_records
 
-    def run(self, *args):
+    def run(self, *args, **kwargs):
         """
         Run handles the generic setup and teardown that happens with
         the running of each step.  The real work that is unique to
@@ -431,6 +443,33 @@ class Step:
             step_result = None
 
             self.log.info("Step %s running with args %s.", self.name, args)
+
+            # Set any uninitialized parameters
+            filename = None
+            if len(args) > 0:
+                filename = args[0]
+
+            try:
+                config, config_file = self.build_config(filename, **kwargs)
+            except NotImplementedError:
+                # Catch steps that cannot build a config
+                # (i.e. post hooks created from local functions)
+                config = {}
+
+            skip = {"class", "logcfg", "name", "config_file"}
+            for key in config:
+                if key in kwargs or (key not in skip and key in self._initialized
+                                     and not self._initialized[key]):
+                    self.log.debug(f'Setting parameter {key} to {config[key]}')
+                    setattr(self, key, config[key])
+                    if key == 'steps':
+                        for step, step_dict in config[key].items():
+                            for step_key, step_value in step_dict.items():
+                                self.log.debug(f'Setting parameters {step}.{step_key} '
+                                               f'to {step_value}')
+                                step_instance = getattr(self, step)
+                                setattr(step_instance, step_key, step_value)
+
             # log Step or Pipeline parameters from top level only
             if self.parent is None:
                 self.log.info(
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -342,7 +342,8 @@ class Step:
         self._input_filename = None
         self._input_dir = None
         self._keywords = kws
-        if _validate_kwds:
+        self._validate_kwds = _validate_kwds
+        if self._validate_kwds:
             spec = self.load_spec_file()
             kws = config_parser.config_from_dict(
                 kws,
@@ -363,10 +364,8 @@ class Step:
         for key, val in kws.items():
             setattr(self, key, val)
 
-            # Mark them as uninitialized, from the user standpoint,
-            # unless they are required to be used as is
-            if _validate_kwds:
-                self._initialized[key] = False
+            # Mark them as uninitialized, from the user standpoint
+            self._initialized[key] = False
 
         # Create a new logger for this step
         self.log = log.getLogger(self.qualified_name)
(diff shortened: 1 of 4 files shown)


melanieclarke:
JWST regtests here:
https://github.com/spacetelescope/RegressionTests/actions/runs/11260804666
Roman regtests here:
https://github.com/spacetelescope/RegressionTests/actions/runs/11261960698

braingram:
The romancal documentation would also need updated for this change: https://roman-pipeline.readthedocs.io/en/latest/roman/stpipe/call_via_run.html

There are uses of `run` in romancal unit tests where the assumption is that the default in-code parameters will be used. With this PR how would one call `run` without fetching CRDS parameters?

Given the scope of the change I think making this opt-in and deprecating the old behavior makes sense providing advice for a user on how to replace `run` with some other call that doesn't fetch parameters.

I gave this PR a brief "test drive" and found one configuration where the parameters from crds are overwriting user-supplied parameters.
```python
from jwst.resample import ResampleStep
pipeline = ResampleStep(weight_type="ivm")
pipeline.run("asn_with_nircam_images.json")
```
on stpipe main I see the expected log message:
```
  weight_type: ivm
```
with this PR I'm seeing:
```
  weight_type: exptime
```
Due to an overwrite from [jwst_nircam_pars-resamplestep_0001.asdf](https://jwst-crds.stsci.edu/browse/jwst_nircam_pars-resamplestep_0001.asdf).
Would you add a unit test that replicates this failure and fixes the issue?

melanieclarke:
Thanks for testing, @braingram.

To disable CRDS pars for run, it currently works the same way call does - if you set the STPIPE_DISABLE_CRDS_STEPPARS environment variable to true, it does not check CRDS.  The command line interface additionally allows a `disable-crds-steppars` argument that will directly disable it as well.  I could add an explicit argument for `run` too, if we want it to behave differently than call.

For input parameters set on instantiation, what you saw is currently expected behavior (expected by me, anyway).  To override parameters that might be in CRDS, with the current code, you would do:
```
from jwst.resample import ResampleStep
pipeline = ResampleStep()
pipeline.run("asn_with_nircam_images.json", weight_type="ivm")
```
or
```
import os
from jwst.resample import ResampleStep
os.environ['STPIPE_DISABLE_CRDS_STEPPARS'] = 'True'
pipeline = ResampleStep(weight_type="ivm")
pipeline.run("asn_with_nircam_images.json")
```

I wrote it this way because there are a number of attributes set on instantiation by default in cases where steps are created by pipelines that should not disable CRDS overrides later, but I will take another look and see if I can distinguish this case from that one.

melanieclarke:
Two changes pushed, @braingram -

Add handling for an explicit disable for CRDS checks in the run call.  Set `disable_crds_steppars` to True if CRDS checks are not desired, False if they're definitely desired.  Default is None, as for call, which checks for the environment variable.

Add initialization handling for parameters explicitly passed on instantiation, for either Step or Pipe, so they are not overridden by CRDS checks.  Your example for ResampleStep should now work as expected.

src/stpipe/step.py
```diff
@@ -431,6 +463,42 @@ def run(self, *args):
...
+            # Get parameters from CRDS
+            if self._validate_kwds:
+                # Get the first filename, if available
+                filename = None
+                if len(args) > 0:
+                    filename = args[0]
+
+                # Build config from CRDS + user keywords
+                try:
+                    parameters, _ = self.build_config(
```
braingram, line 488: Can this be skipped entirely if `disable_crds_steppars==True`? It appears to be opening the input file and doing nothing with it in that case.

src/stpipe/step.py
```diff
@@ -431,6 +463,42 @@ def run(self, *args):
...
+                        filename, disable=disable_crds_steppars, **kwargs
+                    )
+                except (NotImplementedError, FileNotFoundError):
+                    # Catch steps that cannot build a config
+                    # (e.g. post hooks created from local functions,
+                    # missing input files)
+                    self.log.warning(f"Cannot retrieve CRDS keywords for {self.name}.")
```
braingram, line 495: My vote is for making this an exception. If the expectation is that this will fetch CRDS parameters and that fails an error sees more useful than output that is incorrect because it doesn't use those parameters.

src/stpipe/step.py
melanieclarke: Good point. I'll fix that.

src/stpipe/step.py
melanieclarke: Sure, I can make this an exception.  I originally made it a warning because things like post-hooks that aren't proper steps might make it to this clause, but I will make sure that case is separately handled.

melanieclarke:
> My vote is still for making this opt-in given the scope of the change. Is it possible with the config tracking in this PR to warn if `run` is called without crds parameters being fetched?

I'm reluctant to make the CRDS checks off by default because that risks the same situation we currently have, where the users don't know that they have to do something special to get the currently recommended default parameters.  But you are right that this is a significant change, so perhaps we can have a short transition period for a build, where CRDS checks are available but off by default, with a warning.  I'll look into where/how to warn.

> Also, is there any reason to keep `call` with this PR? Isn't it now a less-convenient version of `run`? Given the documentation updates that are needed for this PR I'd say we deprecate call as part of this change so the documentation can be updated to say call is no longer preferred.

Only that it is in very common use now, since we have been recommending it in place of run for years.  If we do want to remove it, we should have a much longer transition period before we do.  If we don't turn on CRDS checks by default in run now, we should wait to deprecate call until we do turn them on.

> One additional question. Are you aware of examples "in the wild" where `run` is used? All the jwst notebooks I can find as part of the spacetelescope repo use `call`.

I think there are still lots out there in less officially vetted places.  Here's a (deprecated) one I wrote, before I was aware of the distinction between run and call:
[caveat example](https://github.com/spacetelescope/jwst-caveat-examples/blob/main/NIRSPEC_General/nrs_recenter_extraction_workaround.ipynb)

src/stpipe/step.py
melanieclarke: Done.  CRDS checks are now always disabled for pre and post hooks.

melanieclarke:
Changes pushed for disabling CRDS checks by default, with a warning.

melanieclarke:
Thanks, I understand your perspective, and I'm also in favor of evaluating the Step API and making more deliberate design choices for the future.  But judging from conversations I have had with users in INS, as well as my own experience in learning to work with the pipeline, the current state of run is what causes confusion - the documentation is at odds with user expectations. Perhaps we could check with INS whether they would prefer to make this change now, while we continue to work on improvements for the future?

PR #195, opened 2024-10-09
standardize setattr use in skip

braingram:
Closes https://github.com/spacetelescope/roman_datamodels/issues/343

jwst regtests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/11804014366
romancal regtests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/11803229018
(it may be worthwhile to add a regression or unit test in romancal for the feature fixed in this PR in a follow-up romancal PR)

This PR standardizes how stpipe sets `meta.cal_step.<class_alias>` when a step is skipped. On main the behavior for a single roman datamodel (not a ModelLibrary) does not log the skip. Running the following on main:

```python
from romancal.pipeline import ExposurePipeline
elp = ExposurePipeline()
elp.saturation.skip = True
elp.refpix.skip = True
elp.linearity.skip = True
elp.dark_current.skip = True
elp.rampfit.skip = True
elp.assign_wcs.skip = True
elp.flatfield.skip = True
elp.photom.skip = True
elp.source_catalog.skip = True
elp.tweakreg.skip = True
result = elp.run("some_uncal.asdf")
print(result[0].meta.cal_step.saturation)
```
results in printing "INCOMPLETE" indicating that even though the saturation step was skipped stpipe failed to log this to the resulting datamodel.

With this PR the above example results in printing "SKIPPED".

## Tasks

</details>

### DIFF
--- a/changes/195.bugfix.rst
+++ b/changes/195.bugfix.rst
@@ -1,0 +1,1 @@
+Fix cal_step setting when a step is skipped for roman datamodels.
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -477,46 +477,28 @@ class Step:
                 # Run the Step-specific code.
                 if self.skip:
                     self.log.info("Step skipped.")
-                    if isinstance(args[0], AbstractModelLibrary):
-                        library = args[0]
-                        with library:
-                            for i, model in enumerate(library):
-                                try:
-                                    setattr(
-                                        model.meta.cal_step, self.class_alias, "SKIPPED"
-                                    )
-                                except AttributeError as e:
-                                    self.log.info(
-                                        "Could not record skip into DataModel "
-                                        "header: %s",
-                                        e,
-                                    )
-                                library.shelve(model, i)
-                    elif isinstance(args[0], AbstractDataModel):
-                        if self.class_alias is not None:
+
+                    if self.class_alias is not None:
+
+                        def set_skipped(model):
+                            try:
+                                setattr(
+                                    model.meta.cal_step, self.class_alias, "SKIPPED"
+                                )
+                            except AttributeError as e:
+                                self.log.info(
+                                    "Could not record skip into DataModel "
+                                    "header: %s",
+                                    e,
+                                )
+
+                        if isinstance(args[0], AbstractModelLibrary):
+                            list(args[0].map_function(lambda m, i: set_skipped(m)))
+                        elif isinstance(args[0], AbstractDataModel):
                             if isinstance(args[0], Sequence):
-                                for model in args[0]:
-                                    try:
-                                        model[f"meta.cal_step.{self.class_alias}"] = (
-                                            "SKIPPED"
-                                        )
-                                    except AttributeError as e:  # noqa: PERF203
-                                        self.log.info(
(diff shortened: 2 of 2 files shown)


braingram:
@schlafly I can't request you as a reviewer but when available would you give this a look (or suggest an alternate roman reviewer)? This PR should fix steps not being marked as "SKIPPED" when skipped view setting the `skip` attribute or by passing `<step>.skip==True` to strun.

schlafly (approved):
Thanks Brett, looks good.  Remind me---I thought we did have some regtests that skipped tests (e.g., for the all saturated test), and I thought you had stumbled on this issue in the process of testing some unrelated code in romancal?  So I had expected some steps to be newly marked SKIPPED.  But perhaps these have all been coded around.

braingram:
> Thanks Brett, looks good. Remind me---I thought we did have some regtests that skipped tests (e.g., for the all saturated test), and I thought you had stumbled on this issue in the process of testing some unrelated code in romancal? So I had expected some steps to be newly marked SKIPPED. But perhaps these have all been coded around.

So far I've only been able to find one test in romancal that passes `skip` to (an equivalent to) `strun`:
https://github.com/spacetelescope/romancal/blob/9031ae12b14b22ba565581670e74b826d40ed86a/romancal/regtest/test_wfi_image_pipeline.py#L283
This test sets `tweakreg.skip=True` however the test checks that `cal_step.tweakreg==INCOMPLETE`. Interestingly this PR doesn't switch the result to `cal_step.tweakreg==SKIPPED` because there's.... yet another bug? The issue is that tweakreg is passed a list of models. stpipe on main (and with this PR, to match main) won't peek into that list to see if it's a list of models (and set the `cal_step` to skipped). I'll give this some more thought but I think this is a separate issue (since tweakreg is passed a list of models, not a model and not a library, both of which are handled in this PR). Part of me thinks stpipe should handle a list of models in a way where the cal_step status is set to skipped if the step is skipped. However this doesn't appear to be the case on main even for jwst as seen in the following
```python
import jwst.datamodels
from jwst.step import RampFitStep

ims = [jwst.datamodels.ImageModel()] * 2
step = RampFitStep()
step.skip = True
result = step.run(ims[0])
print(result.meta.cal_step.ramp_fit)

ims = [jwst.datamodels.ImageModel()] * 2
result = step.run(ims)
print(result[0].meta.cal_step.ramp_fit)
```
This prints out (in addition to all the pipeline logging stuff):
```
SKIPPED
None
```
When passed a model stpipe marks it as skipped when the step is skipped. When passed a list of models, stpipe doesn't mark any models as skipped. This is the same when run with main or with this PR.

For the fully saturated tests (which trigger [this part](https://github.com/spacetelescope/romancal/blob/9031ae12b14b22ba565581670e74b826d40ed86a/romancal/pipeline/exposure_pipeline.py#L153) of the exposure pipeline) the skipping of steps is done within the pipeline (not via stpipe) which also sets the step status. Those uses aren't impacted by this PR.

schlafly:
Ugh.  Thanks.  I agree with you that this PR looks good and should proceed.

nden (approved):
lgtm

PR #197, opened 2024-10-10
Unit tests for step.save and step.skip

emolter:
This PR adds the unit tests from https://github.com/spacetelescope/stpipe/pull/190 as it looks like that PR is likely to be withdrawn, at least for the time being, but the unit tests are still relevant to the main branch.

## Tasks

</details>

### DIFF
--- a/tests/test_step.py
+++ b/tests/test_step.py
@@ -423,12 +423,11 @@ class StepWithModel(Step):
     output_ext = string(default='simplestep')
     save_results = boolean(default=True)
     """
-    # spec = """
-    #
-    # skip = bool(default=False)
-    # """
 
     def process(self, input_model):
+        # make a change to ensure step skip is working
+        # without having to define SimpleDataModel.meta.stepname
+        input_model.stepstatus = "COMPLETED"
         return input_model
 
 
@@ -439,17 +438,12 @@ class SimpleDataModel(AbstractDataModel):
     def crds_observatory(self):
         return "jwst"
 
-    # @property
-    # def meta(self):
-    #     return {"filename": "test.fits"}
-
     def get_crds_parameters(self):
         return {"test": "none"}
 
     def save(self, path, dir_path=None, *args, **kwargs):
         saveid = getattr(self, "saveid", None)
         if saveid is not None:
-            print(f"here {saveid}")
             fname = saveid+"-saved.txt"
             with open(fname, "w") as f:
                 f.write(f"{path}")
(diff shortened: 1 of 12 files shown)


tests/test_step.py
```diff
@@ -411,3 +414,162 @@ def test_log_records():
...
+)
+def test_save_container(tmp_cwd, model_list):
+    """ensure list-like save still works for non-list sequence"""
+    container = SimpleContainer(model_list)
+    step = StepWithModel()
+    step.run(container)
+    for i in range(3):
+        assert (tmp_cwd / f"test{i}-saved.txt").exists()
+
+
```
braingram, line 539:
```suggestion
```

My vote is for making an issue/ticket for this instead of adding a known failing test. For the issue perhaps you would add some details about how this fails (does it throw an exception, overwrite files, not save?).

tests/test_step.py
```diff
@@ -411,3 +414,162 @@ def test_log_records():
...
+def test_skip_container(tmp_cwd, model_list):
```
braingram, line 540: To confirm, there are no unit tests that currently test that `Step.skip` skips a step (for any input)? I'm failing to find any.

emolter: Not that I remember.  This adds one for a model and one for a container.  Do you think I should add one for a library as well? or other types of input?

tests/test_step.py
```diff
@@ -411,3 +414,162 @@ def test_log_records():
...
+
+    def get_crds_parameters(self):
+        return {"test": "none"}
+
+    def save(self, path, dir_path=None, *args, **kwargs):
+        saveid = getattr(self, "saveid", None)
+        if saveid is not None:
+            fname = saveid + "-saved.txt"
+            with open(fname, "w") as f:
+                f.write(f"{path}")
```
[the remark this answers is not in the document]
braingram: Thanks for the clarification. So the idea is that `save` uses a different path so that way there's no chance of a false positive where `save` isn't used by a file of the expected path is produced. Thanks!

tests/test_step.py
braingram: Thanks. I think the tests in this PR are a good starting point. Further tests can go in a future PR.

tests/test_step.py
emolter: https://github.com/spacetelescope/stpipe/issues/218

braingram (approved):
Thanks!

PR #202, opened 2024-10-24
support selecting steps by package name for strun

braingram:
If a user installs both romancal and jwst and runs the following:
```
strun resample some_file
```
strun will always find the resample step in jwst (even though romancal provides a resample step with the class alias "resample").

This PR adds the ability to filter steps found by strun by providing the package name as a "scope. With this PR a user could call:

```
strun romancal::resample some_file
```

to run resample from romancal or

```
strun jwst::resample some_file
```

to run resample from jwst.

Documentation in both packages will need to be updated minimally with a description that providing a package name may be required if multiple strun supported packages are provided.

## Tasks

</details>

### DIFF
--- a/src/stpipe/cli/strun.py
+++ b/src/stpipe/cli/strun.py
@@ -1,6 +1,6 @@
 import sys
 
-from stpipe import Step
+from stpipe import cmdline
 from stpipe.cli.main import _print_versions
 from stpipe.exceptions import StpipeExitException
 
@@ -21,7 +21,7 @@ def main():
         sys.exit(0)
 
     try:
-        Step.from_cmdline(sys.argv[1:])
+        cmdline.step_from_cmdline(sys.argv[1:])
     except StpipeExitException as e:
         sys.exit(e.exit_status)
     except Exception:
--- a/changes/202.feature.rst
+++ b/changes/202.feature.rst
@@ -1,0 +1,1 @@
+Allow class aliases (used during strun) to contain the package name (for example "jwst::resample").
--- a/src/stpipe/entry_points.py
+++ b/src/stpipe/entry_points.py
@@ -1,7 +1,7 @@
 import warnings
 from collections import namedtuple
 
-from importlib_metadata import entry_points
+import importlib_metadata
 
 STEPS_GROUP = "stpipe.steps"
 
@@ -26,7 +26,7 @@ def get_steps():
     """
     steps = []
 
-    for entry_point in entry_points(group=STEPS_GROUP):
+    for entry_point in importlib_metadata.entry_points(group=STEPS_GROUP):
         package_name = entry_point.dist.name
         package_version = entry_point.dist.version
         package_steps = []
--- a/src/stpipe/utilities.py
+++ b/src/stpipe/utilities.py
@@ -33,13 +33,29 @@ def resolve_step_class_alias(name):
     else:
         scope, class_name = None, name
 
+    # track all found steps keyed by package name
+    found_class_names = {}
     for info in entry_points.get_steps():
         if scope and info.package_name != scope:
             continue
         if info.class_alias is not None and class_name == info.class_alias:
-            return info.class_name
+            found_class_names[info.package_name] = info
 
-    return name
+    if not found_class_names:
+        return name
+
+    if len(found_class_names) == 1:
+        return found_class_names.popitem()[1].class_name
+
+    # class alias resolved to several possible steps
+    scopes = list(found_class_names.keys())
+    msg = (
+        f"class alias {name} matched more than 1 step. Please provide "
+        "the package name along with the step name. One of:\n"
+    )
+    for scope in scopes:
+        msg += f"  {scope}::{name}\n"
+    raise ValueError(msg)
 
 
 def import_class(full_name, subclassof=object, config_file=None):
--- a/src/stpipe/utilities.py
+++ b/src/stpipe/utilities.py
@@ -19,13 +19,24 @@ def resolve_step_class_alias(name):
     Parameters
     ----------
     name : str
+        If name contains "::" only the package with
+        a name matching the characters before "::"
+        will be searched for the matching step.
 
     Returns
     -------
     str
     """
+    # check if the name contains a package name
+    if "::" in name:
+        scope, class_name = name.split("::", maxsplit=1)
+    else:
+        scope, class_name = None, name
+
     for info in entry_points.get_steps():
-        if info.class_alias is not None and name == info.class_alias:
+        if scope and info.package_name != scope:
+            continue
+        if info.class_alias is not None and class_name == info.class_alias:
             return info.class_name
 
     return name
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -56,39 +56,99 @@ def test_import_func_no_module():
         import_func("foo")
 
 
-@pytest.mark.parametrize(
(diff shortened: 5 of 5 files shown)


src/stpipe/cli/strun.py
```diff
@@ -21,7 +21,7 @@ def main():
         sys.exit(0)
 
     try:
-        Step.from_cmdline(sys.argv[1:])
+        cmdline.step_from_cmdline(sys.argv[1:])
```
braingram, line 24: This change is not required by this PR (I'm fine removing it) but during debugging it was confusing that this interface calls `Step.from_cmdline` which then just calls `cmdline.step_from_cmdline`.
https://github.com/spacetelescope/stpipe/blob/799be658b5fbd36e1c4c47a07e00f76575b51d97/src/stpipe/step.py#L170-L191

braingram:
Pinging @tapastro and @schlafly for comments on this PR before running regression tests. What do you both think of the new way of disambiguating steps when both jwst and romancal are installed?

zacharyburnett:
the only concern I have is if some user's shell uses `::` as an operator, but in that case they could just surround the argument with quotes

braingram:
> the only concern I have is if some user's shell uses `::` as an operator, but in that case they could just surround the argument with quotes

Good point. Is there a shell that uses that? I don't think bash does and I'm too lazy to try the fancy "new" ones xD

zacharyburnett:
the only shell I can think of that uses colons is `nushell` but if you're using nushell then you're probably quoting all your arguments anyways just to be safe haha

melanieclarke:
Question - isn't it already possible to disambiguate by specifying the package path for the step, e.g. 'strun jwst.resample.ResampleStep'?  Is the idea just to add another convenient alias?

braingram:
I mean maybe we dip into unicode... `strun jwst😍resample`

zacharyburnett:
> Question - isn't it already possible to disambiguate by specifying the package path for the step, e.g. 'strun jwst.resample.ResampleStep'? Is the idea just to add another convenient alias?

I think the intent is to reduce unexpected behavior, if a user forgets they have both installed and just calls `resample`. However, this won't prevent the user from calling `resample` by itself right? So the problem persists.

braingram:
> Question - isn't it already possible to disambiguate by specifying the package path for the step, e.g. 'strun jwst.resample.ResampleStep'? Is the idea just to add another convenient alias?

Yeah it's possible to either use the full step class or the class alias. For jwst these are defined [here](https://github.com/spacetelescope/jwst/blob/a2d6ba8b0a0eface87d6473738aa1c208f0a8e85/jwst/stpipe/integration.py#L78) and for romancal [here](https://github.com/spacetelescope/romancal/blob/07d66cf094e837e6802e71f0d5d435956e4d979f/romancal/stpipe/integration.py#L41). Some "class aliases" (the second item in each tuple) are common which leads to issues when calling `strun resample ...`.

braingram:
> > Question - isn't it already possible to disambiguate by specifying the package path for the step, e.g. 'strun jwst.resample.ResampleStep'? Is the idea just to add another convenient alias?
> 
> I think the intent is to reduce unexpected behavior, if a user forgets they have both installed and just calls `resample`. However, this won't prevent the user from calling `resample` by itself right? So the problem persists.

We could also add a warning or exception if multiple classes are matched. I think that's worth doing (maybe even part of this PR).

To provide a bit more context. The "class alias" is also used for setting the "meta.cal_step" status when the step is skipped.
https://github.com/spacetelescope/stpipe/blob/799be658b5fbd36e1c4c47a07e00f76575b51d97/src/stpipe/step.py#L484-L486
Which is contributing to https://github.com/spacetelescope/roman_datamodels/issues/343 where if a step is skipped in a pipeline for romancal it's cal_step status is not set to "SKIPPED". This is somewhat a separate issue but as romancal is going to define more class aliases to fix the skip issue it seems worth having a way to disambiguate the alias provided to strun since we will soon have more conflicting aliases (and both pipelines document that using an alias is possible).

schlafly:
This seems like a good design to me, thanks!

melanieclarke:
I see.  I thought the original intent was that the class alias should be globally unique for stpipe packages, but it does seem silly to make romancal think of a synonym for "resample".

tapastro:
Agreed that this seems like a reasonable approach - I think a warning or exception is a useful, if not necessary, addition. 

I am curious who the "typical" user would be who would encounter this issue, and in what environment this overlap would occur. I am not that familiar with what a typical Roman user is expected to do wrt. processing, locally vs. in the cloud, but I kind of expected processing for the two observatories to require different environments, if not AWS instances vs. local processing. Not that this should prevent a solution, just some speculation!

schlafly:
I agree that especially as currently Roman isn't operational there aren't many people with both the Webb and Roman pipelines running.  But it seemed bad form for the two packages using stpipe to step on one anothers' toes.  I think we do want people to be able to have both pipelines running side-by-side in the same environment.

nden:
Roman data will be processed locally as well. There's no reason to expect people won't be interested in both JWST and Roman data.

braingram:
> Agreed that this seems like a reasonable approach - I think a warning or exception is a useful, if not necessary, addition.
> 
> I am curious who the "typical" user would be who would encounter this issue, and in what environment this overlap would occur. I am not that familiar with what a typical Roman user is expected to do wrt. processing, locally vs. in the cloud, but I kind of expected processing for the two observatories to require different environments, if not AWS instances vs. local processing. Not that this should prevent a solution, just some speculation!

Thanks! I'll update this PR with an error (and we'll see if we need to dial it back to a warning).

I doubt this will come up for the specific combination of roman and jwst (although @nden objects to saying "you can't have jwst and romancal installed" and I agree). I think the issue is more likely to come up for things like Eureka (which extends the jwst pipeline with additional steps) and as more projects adopt stpipe (we've so far heard of one group working on 1 possibly 2 pipelines).

braingram:
I added an error in [fd68d87](https://github.com/spacetelescope/stpipe/pull/202/commits/fd68d872627eb69f979787c757a67697313b2a2a)

Testing with jwst and romancal installed if I run `strun resample ...` I get:
```
ValueError: class alias resample matched more than 1 step. Please provide the package name along with the step name. One of:
  jwst::resample
  romancal::resample
```
With that environment I can run `strun jwst::resample ...` and it finds and executes the jwst resample step and run `strun romancal::resample ...` and it finds and executes the romancal resample step.

zacharyburnett (approved):
this looks like a good solution to me, and scaleable to more packages in the future. I think it should remain an error, to enforce muscle habit of specifying the package name before shared aliases. To that end, should all aliases require the package name?

braingram:
> this looks like a good solution to me, and scaleable to more packages in the future. I think it should remain an error, to enforce muscle habit of specifying the package name before shared aliases. To that end, should all aliases require the package name?

Thanks! I'll run regtests to make sure the error isn't a problem (although I doubt anything will fail given the test setups).

I'm ok leaving aliases out when strun can unambiguously sort out the class name. If this goes in I'll work on PRs for jwst and romancal RTD to mention the need for a package name if >1 stpipe supporting package provides the same step names.

braingram:
Regtests:
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/11522120325
romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/11522122685

PR #229, opened 2025-03-21
add Step._get_crds_parameters

### LINKED ISSUE #228: Update uses of `get_crds_parameters` to allow for more efficient file opening
`get_crds_parameters` returns a subset of metadata that's passed to crds for reference file selection. Since only a subset is needed if `Step` and `Pipeline` are processing an input file (and not an input `DataModel`) it may be possible to perform a partial read of the file to more efficiently access the same information. This issue is to layout API changes needed to accommodate this mode.

The current `get_crds_parameters` uses are:
In `Step.get_reference_file`:
https://github.com/spacetelescope/stpipe/blob/9692bfaf414010258e7b4fa6c752091d61bd9bfa/src/stpipe/step.py#L809-L814
The use of `open_model` here seems superfluous as the input should be an open `DataModel` (and `open_model` will sometimes copy the input). Within the scope of this ticket I think this usage can be ignored.

In `Step.get_config_from_reference`:
https://github.com/spacetelescope/stpipe/blob/9692bfaf414010258e7b4fa6c752091d61bd9bfa/src/stpipe/step.py#L862-L867
This is likely to receive as input a filename instead of a model due to usage in [cmdline](https://github.com/spacetelescope/stpipe/blob/9692bfaf414010258e7b4fa6c752091d61bd9bfa/src/stpipe/cmdline.py#L317).

In `Pipeline.get_config_from_reference`:
https://github.com/spacetelescope/stpipe/blob/9692bfaf414010258e7b4fa6c752091d61bd9bfa/src/stpipe/pipeline.py#L187-L193
This looks functionally the same as the usage in `Step.get_config_from_reference`.

In `Pipeline._precache_references_impl`:
https://github.com/spacetelescope/stpipe/blob/9692bfaf414010258e7b4fa6c752091d61bd9bfa/src/stpipe/pipeline.py#L319-L321
This is indirectly called from `Pipeline._precache_references`:
https://github.com/spacetelescope/stpipe/blob/9692bfaf414010258e7b4fa6c752091d61bd9bfa/src/stpipe/pipeline.py#L263-L266
which calls `Pipeline._precache_references_opened` which [iterates through any container](https://github.com/spacetelescope/stpipe/blob/9692bfaf414010258e7b4fa6c752091d61bd9bfa/src/stpipe/pipeline.py#L281-L292).

braingram:
Closes #228
Partially resolves [JP-3949](https://jira.stsci.edu/browse/JP-3949)

This PR adds a class method `Step._get_crds_parameters` to allow subclasses to control how files/models are handled for determining crds parameters for fetching step configurations and reference files.

Combined with https://github.com/spacetelescope/stdatamodels/pull/445 this will allow jwst to more efficiently load models when only the metadata is needed (as is the case for crds).

Regtests all pass:
romancal regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/14038435767
jwst regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/14038446114

## Tasks

</details>

### DIFF
--- a/src/stpipe/pipeline.py
+++ b/src/stpipe/pipeline.py
@@ -182,7 +182,7 @@ class Pipeline(Step):
         logger.debug("Retrieving all substep parameters from CRDS")
         #
         # Iterate over the steps in the pipeline
-        _, crds_parameters, crds_observatory = cls._get_crds_parameters(dataset)
+        crds_parameters, crds_observatory = cls._get_crds_parameters(dataset)
 
         for cal_step in cls.step_defs.keys():
             cal_step_class = cls.step_defs[cal_step]
@@ -252,9 +252,7 @@ class Pipeline(Step):
         None
         """
         try:
-            filename, crds_parameters, observatory = self._get_crds_parameters(
-                input_file
-            )
+            crds_parameters, observatory = self._get_crds_parameters(input_file)
         except (ValueError, TypeError, OSError):
             self.log.info("First argument %s does not appear to be a model", input_file)
             return
@@ -269,7 +267,7 @@ class Pipeline(Step):
 
         self.log.info(
             "Prefetching reference files for dataset: %r reftypes = %r",
-            filename,
+            self._get_filename(input_file),
             fetch_types,
         )
         crds_refs = crds_client.get_multiple_reference_paths(
--- a/src/stpipe/pipeline.py
+++ b/src/stpipe/pipeline.py
@@ -2,14 +2,12 @@
 Pipeline
 """
 
-from collections.abc import Sequence
 from os.path import dirname, join
 from typing import ClassVar
 
 from astropy.extern.configobj.configobj import ConfigObj, Section
 
 from . import config_parser, crds_client, log
-from .library import AbstractModelLibrary
 from .step import Step, get_disable_crds_steppars
 from .utilities import _not_set
 
@@ -184,13 +182,7 @@ class Pipeline(Step):
         logger.debug("Retrieving all substep parameters from CRDS")
         #
         # Iterate over the steps in the pipeline
-        with cls._datamodels_open(dataset, asn_n_members=1) as model:
-            if isinstance(model, Sequence):
-                crds_parameters = model._models[0].get_crds_parameters()
-                crds_observatory = model.crds_observatory
-            else:
-                crds_parameters = model.get_crds_parameters()
-                crds_observatory = model.crds_observatory
+        _, crds_parameters, crds_observatory = cls._get_crds_parameters(dataset)
 
         for cal_step in cls.step_defs.keys():
             cal_step_class = cls.step_defs[cal_step]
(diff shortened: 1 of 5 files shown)


src/stpipe/pipeline.py
```diff
@@ -184,13 +182,7 @@ def get_config_from_reference(cls, dataset, disable=None, crds_observatory=None)
...
-            if isinstance(model, Sequence):
-                crds_parameters = model._models[0].get_crds_parameters()
-                crds_observatory = model.crds_observatory
-            else:
-                crds_parameters = model.get_crds_parameters()
-                crds_observatory = model.crds_observatory
+        crds_parameters, crds_observatory = cls._get_crds_parameters(dataset)
 
         for cal_step in cls.step_defs.keys():
             cal_step_class = cls.step_defs[cal_step]
```
emolter, line 188: I'm testing this out using a real override for the jwst implementation of stpipe.step.  I noticed something that I think is "sort of" a bug:

The present implementation of `_get_crds_parameters()` can handle receiving its own output `parameters` dictionary as input.  But it does so in a way that was hidden to me: it opens the dictionary as a datamodel, then calls itself again so that `model.get_crds_parameters()` is called on it.

On lines 185-189 here, this happens in practice: `_get_crds_parameters` is called with `dataset` of type `dict` in the case that a pipeline is called as a step of another pipeline.  This happens for jwst's `MasterBackgroundMosStep`.  For regular Steps, this is handled on [this line](https://github.com/braingram/stpipe/blob/585b6537c3af53a3360d1fdf6042f9a0329dd544/src/stpipe/step.py#L927) of `Step.get_config_from_reference` but no such bypass exists here in the Pipeline version of `get_config_from_reference`.

This could be handled in the jwst override of `_get_crds_parameters()` by adding explicit handling of dictionaries.  However, (1) it seems a bit strange to handle dicts there and (2) the base implementation performs some datamodel opening and other manipulation here just to get back exactly what it started with which is arguably not the most efficient or easy to understand.

braingram: What condition results in `_get_crds_parameters` receiving a `dataset` that is a `dict`? Would you share an example, point me to an example or code?

Looking at the docstrings for the functions that call `_get_crds_parameters` I don't see any that list `dict` as a supported input.

emolter: It happens for e.g. the jwst unit test `pipeline/tests/test_calwebb_spec2.py`.  Using jwst@main and your stpipe version (although I think there might be some weirdness on main too), and adding the lines
```
        if isinstance(dataset, dict):
            raise TypeError(
                "dataset must be a model, not a dictionary"
            )
```
just above the call, the error is raised when `cls` is `<class 'jwst.master_background.master_background_mos_step.MasterBackgroundMosStep'>`

braingram: I think I see what you're describing. Let me know if I have this right.

On stpipe main a call to [Pipeline.get_config_from_reference](https://github.com/braingram/stpipe/blob/370e10403acd602470491c9452781142fb757681/src/stpipe/pipeline.py#L145) will open the provided `dataset` and call [get_crds_parameters](https://github.com/braingram/stpipe/blob/370e10403acd602470491c9452781142fb757681/src/stpipe/pipeline.py#L187) assigning to a variable `crds_parameters` which is later passed to `get_config_from_reference` for each [step of the pipeline](https://github.com/braingram/stpipe/blob/370e10403acd602470491c9452781142fb757681/src/stpipe/pipeline.py#L197C56-L197C81). If a `Pipeline` instance is used as a `Step` (which for some reason is the case for [MasterBackgroundMosStep](https://github.com/spacetelescope/jwst/blob/6cb593526baa0f83b6ec62fcba1321b00b6cd5b9/jwst/pipeline/calwebb_spec2.py#L80) in `calwebb_spec2`) or even for `Step` instances that are parts of a `Pipeline` this results in providing `get_config_from_reference` a non-supported input (`dict` is not listed as supported for either method).

So currently on stpipe main for `MasterBackgroundMosStep` the input `dict` is being opened as a model then `get_crds_parameters` is called on that model.

Is that right so far in describing the current behavior on main? That does sound like a bug to me and maybe `get_config_from_reference` should accept a `dict` as an input? 

With this PR the call to `Pipeline.get_config_from_reference` results in the input dataset being passed to `_get_crds_parameters`. The returned dict is assigned to `crds_parameters` which is later passed to `get_config_from_reference`. Isn't this essentially the same as what is happening on main?

If the above is correct, are you suggesting we try to fix that bug in this PR or is there some other bug introduced in this PR that I'm not understanding?

emolter: Yes this is exactly what I'm describing.  I agree with you, this bug is not introduced by the PR and also exists on main.  Sorry for that confusion.  I wanted to make sure I understood this properly myself and that your PR was indeed not the culprit.

As far as whether to fix it here or just to open an issue, I guess I don't care at the end of the day because there's a workaround for what I want to do with jwst.  I do think fixing it would make the subclasses make more sense because they wouldn't have to handle dicts.  I think we're once again running into the "there aren't enough active stpipe maintainers" problem... normally I'd say it should be handled separately but if we don't handle it here I'm concerned it'd never get fixed.

braingram: I added `dict` as a supported input in https://github.com/spacetelescope/stpipe/pull/229/commits/0add84447f203848f03c50c38e0000ee214314f4

Would you give that a try?

I will restart the regression tests.

emolter (approved):
Looks good to me, thanks Brett.  As far as I can tell, this allows me to override `get_crds_parameters` in the way we want for jwst.

I'm not too familiar with `stpipe` yet.  I think it'd be good to get another reviewer, not sure who that would be though.

braingram:
@schlafly would you give this a review? I can't request you as a reviewer.

tapastro (approved):
Nothing caught my eye - if tests pass with this implementation, 👍 from me.

schlafly (approved):

emolter:
> Nothing caught my eye - if tests pass with this implementation, 👍 from me.

The regression tests below show that JWST can indeed use this for the CRDS queries:
https://github.com/spacetelescope/RegressionTests/actions/runs/14783074620

PR #238, opened 2025-06-16
WIP: Refactor stpipe logging

melanieclarke:
Resolves [JP-1866](https://jira.stsci.edu/browse/JP-1866)
Resolves [JP-4008](https://jira.stsci.edu/browse/JP-4008)

Draft proposal to simplify and consolidate logging under a single "stpipe" root logger.

This will need discussion and coordination across romancal and jwst teams, as it is currently a breaking change.  To demonstrate the changes needed, I filed draft PRs for jwst, stdatamodels, and stcal here:
https://github.com/spacetelescope/jwst/pull/9547
https://github.com/spacetelescope/stdatamodels/pull/499
https://github.com/spacetelescope/stcal/pull/377

If we like this proposal, these repositories will also need similar changes before this PR can be merged:
romancal
roman_datamodels

Note:

I intended this PR to be a platform for discussion for the logging refactor, but there are a couple other concurrent ideas about logging and it’s not clear at this point if the work will continue here or elsewhere.

In case it helps the discussion or future work, these are the design principles I was working toward here.

Goal: align ST calibration pipeline logging with Python best practices
Principles:
- Nothing happens on import other than getting a logger by name
   - To address this, the delegation logging needs to be removed from stpipe.
   - Downstream packages should not use setLevel or addHandler after getting a logger by name.
- Log handlers are only added at the application level.
   - For stpipe, this means handlers should only be added in the command line script.
   - Since this would be a surprising change to users, INS requests that a stream handler is also optionally added in the call function.  It should be removed when the call process is complete.
- Users should be able to configure logging for calibration in their Python code using only the standard logging module.
   - For stpipe, this means removing the `logcfg` arguments to command line and call/run.
   - For the `call` function, the default option to add a stream handler should not override any previous configuration set by the user.
- Users should be able to easily configure the log from the command line script.
   - For stpipe, replace the logcfg file with simpler command line arguments to control the log level and format and optionally write to a log file.
- Remove the self.log anti-pattern.
   - In stpipe, remove the step.log attribute.
   - In downstream packages, use getLogger to get logs by name and log to them instead of self.log.
- Don’t add handlers to the root logger. 
   - In stpipe, this is also addressed by removing the delegation logging.
   - In downstream packages, always use loggers that start with a known parent name so that stpipe can manage their configuration.
- Preferably provide a single entry point for configuring the log for all packages in the ST system.
   - In downstream packages, log all messages to loggers that start with the name “stpipe”.

The last principle I think is optional, but I felt would be nice for maximum utility to users.  If all cal packages agree to log to the same parent log name, then all a user needs to do to configure all logs from all our complex interdependent packages is to set it up once on the "stpipe" parent log.  If it is objectionable to log to an stpipe logger from a non-stpipe package, we could all agree on a different parent name, e.g. "stlog" – the actual name of the log is arbitrary.  Alternately, if we do not want to have all packages log to the same logger, we could also consider white-listing known log names, either through entry points or some other configuration local to the jwst and romancal packages, so that stpipe knows which ones it should handle.  The option implemented here is currently the simplest one: stpipe configures only one log, called "stpipe".

## Tasks

</details>


src/stpipe/step.py
```diff
@@ -735,12 +721,15 @@ def call(cls, *args, **kwargs):
...
-        a new instance but simply runs the existing instance of the `Step`
-        class.
         """
+        # Check for an existing log format for stpipe and initialize
+        # with a default setting if not already done
+        configure_log = kwargs.pop("configure_log", True)
+        stpipe_log = logging.getLogger("stpipe")
+        if configure_log:
+            if len(stpipe_log.handlers) == 0:
+                log_config.set_log_configuration()
```
melanieclarke, line 731: Note for self-review: this should use a context manager to add the stream handler only for the duration of the function.

melanieclarke:
I will close this prototype PR to start fresh from the current state in stpipe, following the merge of #240.

From various discussions, it sounds like there is a strong desire to make sure the standard convention of getting a local logger with `logging.getLogger(__name__)` continues to work in the downstream packages.  In the next prototype, I'll look into whitelisting expected loggers instead of configuring only one log.

There is also significant concern about the impact on downstream packages, so I will prioritize incremental improvements and appropriate deprecations in the next version.

PR #240, opened 2025-06-16
WIP: remove delegation

braingram:
Resolves [JP-nnnn](https://jira.stsci.edu/browse/JP-nnnn)

Closes #

This PR addresses ...

## Tasks

</details>


braingram:
I don't know why I can't request reviewers but pinging @melanieclarke @tapastro @schlafly for comments/testing/reviews.

tapastro:
I tested this with a custom log config file and with the cal_logs attribute, just to check that both are still behaving properly. It looks good from my point of view.

melanieclarke:
This looks like a reasonable alternate partial refactor for the log, if we want to make sure 3.13.4 and 3.13.5 are covered now.

I will note that it changes the behavior for the root logger, so it may surprise downstream users.  For example:
```
import stpipe
import logging
log = logging.getLogger()
log.info('test')
```
will print `2025-06-17 16:17:58,159 - root - INFO - test` on this branch because a stream handler is added to the root logger with level INFO when stpipe is imported.  With the previous delegator solution, no message was printed.  

I expect this change will probably lead to things like duplicate log messages for any stpipe user that calls any other code that adds log handlers.  I know the stpipe log already interacts poorly with other packages that use logging, but this change may make it worse in some situations.

melanieclarke:
One more note -- there is significant overlap here with #238, which also removes the delegation logging, in favor of never touching the root logger and only configuring the "stpipe" logger at the application level.  This is intended to avoid unexpected issues like the above.  I would personally prefer to defer log refactoring to next build, so we have time to test, agree on a solution, and make sure we're not making things worse.

braingram:
> This looks like a reasonable alternate partial refactor for the log, if we want to make sure 3.13.4 and 3.13.5 are covered now.
> 
> I will note that it changes the behavior for the root logger, so it may surprise downstream users. For example:
> 
> ```
> import stpipe
> import logging
> log = logging.getLogger()
> log.info('test')
> ```
> 
> will print `2025-06-17 16:17:58,159 - root - INFO - test` on this branch because a stream handler is added to the root logger with level INFO when stpipe is imported. With the previous delegator solution, no message was printed.

Thanks. I pushed a few commits that test the example you shared and avoid this modification (outside of where it's necessary to capture/format/display the logs).

> I expect this change will probably lead to things like duplicate log messages for any stpipe user that calls any other code that adds log handlers. I know the stpipe log already interacts poorly with other packages that use logging, but this change may make it worse in some situations.

For the situation where a logger is already configured with a handler (let's say logger 'foo' has a stream handler to display to stderr) than with this PR (as on main) the log messages emitted during a run will appear twice in stderr. Isn't this the intended result from that setup? Presumably the handler added to the 'foo' logger was added to display those messages in a certain format (as is also the case for the stpipe added handler).

melanieclarke:
Thanks for your efforts, but I would still very much prefer to wait until next build to tackle log refactoring for stpipe, rather than continuing to add complexity to the workaround here in the last few days of this build.

braingram:
I'm marking this as draft since pinning is the plan for this build. We can discuss these options after.

src/stpipe/cmdline.py
```diff
@@ -156,7 +159,7 @@ def set_value(subconf, key, val):
             set_value(config, key, val)
 
 
-def just_the_step_from_cmdline(args, cls=None):
+def just_the_step_from_cmdline(args, cls=None, apply_log_cfg=False):
```
braingram, line 162: This causes 1 jwst unit test error:
```
FAILED jwst/jwst/stpipe/tests/test_step.py::test_disable_crds_steppars_cmdline[--verbose-None-<lambda>] - AssertionError: assert False
 +  where False = <function <lambda> at 0x7f62350cc9a0>('')
 +    where '' = CaptureResult(out='', err='').err
```
and was added to limit the scope of modification of the root logger. The now failing jwst unit test is (as far as I can tell) the only use of `just_the_step_from_cmdline`. The test looks like it would fit into stpipe so could be moved to stpipe (and fixed). An alternative would be to strip the limited scope of the root logger modifications out of this PR (which I think would greatly simplify this PR and more closely match current stpipe where the delegation handler is added to the root logger on import).

braingram:
Re-opened for review. I haven't yet started on the corresponding jwst PR (see https://github.com/spacetelescope/stpipe/pull/240#discussion_r2175181321) but that will need to go along with this.

braingram:
Summary of logging tests using the romancal regtest suite is here: https://innerspace.stsci.edu/pages/viewpage.action?pageId=601760253

src/stpipe/cmdline.py
braingram: https://github.com/spacetelescope/jwst/pull/9616 fixes the jwst unit test (by switching it to use documented API `Step.from_cmdline`). Longer term moving this (and possibly other stpipe-specific tests) to stpipe makes sense but that can be handled in follow-up PRs.

WilliamJamieson (approved):
These changes look good to me.

Note, I actually prefer the change in log message as it gives more information on what the message is and where it came from. It appears this `DelegationHandler` is just stripping information from the log and IMOP that information should not be stripped.

braingram:
Fresh round of regtests (with default python versions)

Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/16171535848
JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/16171538482

schlafly (approved):
Thanks Brett!

tapastro (approved):
👏

PR #255, opened 2025-08-04
Allow downstream packages to specify known loggers

### LINKED ISSUE #245: Allow specifying known loggers instead of configuring root logger
Currently, stpipe attaches handlers to the root logger to configure and capture log messages from downstream packages.  It would be preferable for better control and security, as well as less impact on Python users, to attach configuration to known loggers only.

This could be accomplished by downstream pipeline packages specifying which loggers they wish to configure.

melanieclarke:
Partially resolves [JP-1866](https://jira.stsci.edu/browse/JP-1866)

Closes #245

Allow downstream packages to specify known loggers, to ensure all desired logging messages are captured without adding configuration to the root logger.

For jwst and romancal, the specific loggers desired can be specified by overriding the get_known_loggers method in JwstStep and RomanStep, respectively.

To allow downstream packages time to transition, the get_known_loggers method in the base class raises a DeprecationWarning and returns the root logger. This should behave identically to the current method, which always attaches configuration to the root logger. In the future, the base class should transition to returning only the "stpipe" logger; at that point, the DeprecationWarning can be removed.

No changes to downstream packages are required before merging, but JWST unit and regression tests will fail due to the DeprecationWarning until a PR is merged to override the get_known_loggers method. PR is: https://github.com/spacetelescope/jwst/pull/9720. It can be merged first, since it just adds a new method to the JwstStep class.

## Tasks
  - [x] run regression tests with this branch installed (`"git+https://github.com/<fork>/stpipe@<branch>"`)

</details>

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -512,7 +512,7 @@ class Step:
         gc.collect()
 
         with log.record_logs(
-            log_names=self.get_known_loggers(), formatter=self._log_records_formatter
+            log_names=self.get_stpipe_loggers(), formatter=self._log_records_formatter
         ) as log_records:
             self._log_records = log_records
 
@@ -756,7 +756,7 @@ class Step:
             log_cfg = None
         ctx = nullcontext if log_cfg is None else log_cfg.context
 
-        log_names = cls.get_known_loggers()
+        log_names = cls.get_stpipe_loggers()
         with ctx(log_names):
             config, config_file = cls.build_config(filename, **kwargs)
 
@@ -989,7 +989,7 @@ class Step:
         return config_parser.ConfigObj()
 
     @staticmethod
-    def get_known_loggers():
+    def get_stpipe_loggers():
         """
         Get the names of loggers to configure.
 
@@ -1000,20 +1000,20 @@ class Step:
 
         Returns
         -------
-        loggers : list
-            List of log names to configure.
+        loggers : tuple of str
+            Tuple of log names to configure.
         """
         # Raising a deprecation warning is deferred until the next
         # release, for easier integration.
 
         # msg = (
-        #     "The default for `get_known_loggers` is currently the root logger. "
+        #     "The default for `get_stpipe_loggers` is currently the root logger. "
         #     "This method should be overridden in downstream packages. "
         #     "The default logger will change to 'stpipe' only in future builds."
         # )
         # warnings.warn(msg, DeprecationWarning, stacklevel=2)
 
-        return [""]
+        return ("root",)
 
     def set_primary_input(self, obj, exclusive=True):
         """
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -512,7 +512,9 @@ class Step:
         """
         gc.collect()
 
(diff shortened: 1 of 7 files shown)


melanieclarke:
Downstream jwst tests are failing because of the DeprecationWarning.

Initial JWST regression tests here, with jwst PR to override the get_known_loggers method (https://github.com/spacetelescope/jwst/pull/9643):
https://github.com/spacetelescope/RegressionTests/actions/runs/16275777394
Tests are passing.

Initial romancal tests here, no override:
https://github.com/spacetelescope/RegressionTests/actions/runs/16293019786
Tests are passing.

src/stpipe/step.py
```diff
@@ -973,6 +977,29 @@ def get_config_from_reference(cls, dataset, disable=None, crds_observatory=None)
...
+        -------
+        loggers : list
+            List of log names to configure.
+        """
+        msg = (
+            "The default for `get_known_loggers` is currently the root logger. "
+            "This method should be overridden in downstream packages. "
+            "The default logger will change to 'stpipe' only in future builds."
+        )
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
braingram, line 1000: To make this change easier to integrate without breaking tests can we defer this deprecation warning to the next release. That way we can:
- add the feature `get_known_loggers` (defaulting to root) (no test failures)
- make a stpipe release (no test failures)
- update jwst and romancal to use the new feature
- deprecate the old behavior

melanieclarke: Sounds reasonable to me.

melanieclarke: Done in [eec135d](https://github.com/spacetelescope/stpipe/pull/255/commits/eec135d59f202840ccf8d2c26f5773747811205b).  I will rerun regression tests on main for jwst and romancal.

tapastro (approved):

tapastro:
@melanieclarke I see some downstream test failures for jwst - are the missing log files expected?

melanieclarke:
> > @melanieclarke I see some downstream test failures for jwst - are the missing log files expected?
> 
> No, that's new. Thanks for pointing that out -- I'll look into it.

This is because of a difference in [how setLevel works](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel) for the root logger vs. a named logger.  For the root logger, setting the level to NOTSET causes it to process all messages.  For a named logger, it defers to the parent (which is the root logger, which defaults to WARNING if not otherwise set).  In the failing jwst unit test, the log configuration sets a file name but not a level, so the level defaulted to WARNING, and no warning messages were emitted so the file was not created.

I think for consistent behavior across root and non-root loggers, it would be best to specify a default level for the logging configuration, rather than using NOTSET.  I pushed up a change to default to INFO, which should fix the jwst downstream tests.

@braingram - this is a change in behavior since you last reviewed, so please let me know if you disagree with the solution.

tapastro:
Kicked off a fresh one here: https://github.com/spacetelescope/RegressionTests/actions/runs/17773359637

melanieclarke:
Romancal tests are passing, so I will go ahead and merge now.  Thanks for reviewing, all!

PR #270, opened 2025-09-23
Remove import level warning capture in log configuration

### LINKED ISSUE #269: Importing stpipe silences all warnings.
```python
import warnings
warnings.warn("test", UserWarning)
```
results in:
```
UserWarning: test
  warnings.warn("test", UserWarning)
```
importing stpipe makes this warning (and I believe all other warnings) silent.
```python
import warnings
import stpipe
warnings.warn("test", UserWarning)
```
produces no output.

melanieclarke:
Partially resolves [JP-4138](https://jira.stsci.edu/browse/JP-4138)

Closes #269 

The `stpipe.log` module currently contains an import-level setting to capture warning messages and direct them to the logging system.  If the `py.warnings` log is not configured, this will hide warning messages from the user.  This PR removes this import level behavior.

## Tasks

</details>

### DIFF
--- a/changes/270.bugfix.rst
+++ b/changes/270.bugfix.rst
@@ -1,0 +1,1 @@
+Remove ``logging.captureWarnings`` setting from the ``stpipe.log`` module.
--- a/src/stpipe/log.py
+++ b/src/stpipe/log.py
@@ -337,6 +337,3 @@ def record_logs(log_names, level=logging.NOTSET, formatter=None):
             for log_name in log_names:
                 logger = logging.getLogger(log_name)
                 logger.removeHandler(handler)
-
-
-logging.captureWarnings(True)


melanieclarke:
Another option would be to catch warnings in the pipeline where they are expected and then log them instead if they are important to capture.

I don't love the idea of capturing warnings during `run` because that would again require configuring the py.warnings logger to get them to show up in the log.

melanieclarke:
Either way, maybe we can keep the scope of this PR limited to the currently proposed change? I think we can at least agree that we should remove the import-level warnings capture here.  If you'd like to propose another solution in the future, you're welcome to.

melanieclarke:
jwst regtests:
https://github.com/spacetelescope/RegressionTests/actions/runs/17980031603

romancal regtests:
https://github.com/spacetelescope/RegressionTests/actions/runs/17980160490

tapastro:
I think the highest priority is that we have a record of any warnings emitted during processing. If captureWarnings is False, any warnings.warn() messages will go to stderr but not to our recorded logs. From my point of view, that means I'd rather have a state where we have undesirable behavior in a post-import, post-process interactive session if it means we log all warnings. This alternative means warnings will not go unseen, but will only be present in stderr. We can additionally copy internal (jwst/stcal/romancal) warnings.warn() messages to log.warning() messages, but we would lose warnings from e.g. astropy, numpy, etc.

With that said, I'm going to open a PR that attempts to dynamically capturewarnings as part of the application of stpipe's configuration. There may be an issue we will discover, in which case my preferred fallback would be to captureWarnings(True) on import, add py.warnings to the loggers list and deal with the interactive session consequences.

Pinging @schlafly for any thoughts from Roman.

PR #301, opened 2026-03-03
Check for N/A in override reference files

melanieclarke:
Closes #299

Check for the special value "N/A" before recording override reference files.  "N/A" should be stored as is; any other non-empty string should be recorded as an absolute path.

## Tasks


braingram (approved):
Thanks for fixing this! LGTM.

PR #303, opened 2026-03-24
Move gc.collect to end of step run

### LINKED ISSUE #172: Move `gc.collect` call to end of `Step.run`
`Step.run` calls `gc.collect` near the beginning of the function body:
https://github.com/spacetelescope/stpipe/blob/9952c365458fc3f01839a912b62ab6c637dc9179/src/stpipe/step.py#L418
As it's likely that `Step.process` will result in allocations, it may be more useful to move this call to after `Step.process`.

pllim:
Closes https://github.com/spacetelescope/stpipe/issues/172

## Tasks
  - [ ] run regression tests with this branch installed (`stpipe@git+https://github.com/pllim/stpipe.git@mv-collect`)
    - [x] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/23512447498 (passed)
    - [x] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/23512476825 (passed)

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -488,8 +488,6 @@ class Step:
         the running of each step.  The real work that is unique to
         each step type is done in the `process` method.
         """
-        gc.collect()
-
         with log.record_logs(
             log_names=self.get_stpipe_loggers(), formatter=self._log_records_formatter
         ) as log_records:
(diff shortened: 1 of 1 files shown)


pllim:
If no one knows for sure and it is not broken, should we close both this PR and the issue as won't fix?

braingram:
Closing these sounds good to me since the tests you ran here didn't indicate any memory savings.

PR #304, opened 2026-03-25
JP-4129: Capture pre-pipeline CRDS checks in ASDF log

pllim:
Resolves [JP-4129](https://jira.stsci.edu/browse/JP-4129)

This PR addresses https://github.com/spacetelescope/jwst/issues/9864

Close #309

Close https://github.com/spacetelescope/jwst/pull/10437

## Tasks
  - [x] run regression tests with this branch installed (`stpipe@git+https://github.com/pllim/stpipe.git@log-all-the-things`)
    - [x] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/23520602776 (passed)
    - [x] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/23520609993 (passed)

### DIFF
--- a/changes/304.bugfix.rst
+++ b/changes/304.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed some step log messages not getting saved to output model logs.
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -76,6 +76,8 @@ class Step:
     # log_records to be saved.
     _log_records_formatter = None
 
+    _prerun_logs = []
+
     @classmethod
     def get_config_reftype(cls):
         """
@@ -128,6 +130,7 @@ class Step:
         name : str, optional
             If provided, use that name for the returned instance.
             If not provided, the following are tried (in order):
+
             - The ``name`` parameter in the config file
             - The filename of the config file
             - The name of returned class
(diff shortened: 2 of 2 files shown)


braingram:
Thanks for working on this. Running roman elp on an uncal file (as an initial test of this) I do see the logs recorded to the datamodel now contain "ExposurePipeline instance created." and other similar log messages that don't appear in the datamodel logs on main.

However the changes are causing many duplicate log messages.

If I run `strun roman_elp some_uncal.asdf` on main I get 1 log message "FlatFieldStep instance created." (as one example).
With this PR that log line shows up 14 times.

pllim:
> shows up 14 times

Hmm do you know where the 14 comes from? Admittedly this isn't the cleanest solution. I have also considered the following but they are both breaking changes.

1. Move where the log capture context manager gets started (maybe in call instead of run).
2. Turn the class methods into instance methods. Why are they class methods to begin with?

I also thought about this but then my brain started hurting.

3. Dive into rabbit hole of injecting MemoryHandler into the mix, but it can be triggered in ways we might not expect.

braingram:
Superseded by #309

PR #306, opened 2026-03-26
[WIP] apply new cookiecutter template from `stsci-package-template`

zacharyburnett:
using the template from https://github.com/spacetelescope/stsci-package-template/pull/36

## Tasks


braingram:
CI is failing for a variety of reasons.

There are a lot of changes in this PR. Do we need to do them all at once or can the template be incrementally applied?

.github/workflows/build.yml
```diff
@@ -54,6 +54,3 @@ jobs:
           subject-path: "dist/*"
       # To upload to PyPI without a token, add this workflow file as a Trusted Publisher in the project settings on the PyPI website
       - uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
-        # TODO remove this when we verify a Trusted Publisher config exists on PyPI
```
braingram, line 57: Does this mean trusted publishing is enabled for this package?

.github/workflows/changelog.yml
```diff
@@ -10,30 +10,23 @@ on:
...
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
-          fetch-depth: 0
-      - run: pip install .
-      - run: pip install towncrier
+          python-version: "3"
+      - run: pip install . towncrier
       - run: towncrier check
-      - run: towncrier build --draft | grep -P '#${{ github.event.number }}'
-  changelog-prevent-manual-edit:
```
braingram, line 31: This is removed? That seems ok given that most contributors are familiar with towncrier fragments.

.github/workflows/label.yml
braingram, line 1: I don't see auto-labeling PRs as beneficial (personally). Is there sufficient benefit to adding:
- another workflow (and dependencies that come with it)
- another configuration file
- another set of categorization beyond the changelog fragment types

.github/workflows/tests.yml
```diff
@@ -1,46 +1,77 @@
...
 on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
   push:
     branches:
       - main
-    tags:
```
braingram, line 7: Why remove running this on tags (which will run on releases)?

.github/workflows/tests.yml
```diff
@@ -1,46 +1,77 @@
...
-      - "*"
-  pull_request:
-  schedule:
-    # Weekly Monday 9AM build
-    # * is a special character in YAML so you have to quote this string
-    - cron: "0 9 * * 1"
```
braingram, line 13: Where are the scheduled runs configured?

.github/workflows/tests.yml
```diff
@@ -1,46 +1,77 @@
...
+        - name: run tests with coverage (latest supported Python, ubuntu-latest)
+          linux: py3-cov
+          python-version: ${{ needs.supported-pythons.outputs.latest }}
+          pytest-results-summary: true
+          coverage: github
+        - name: run tests (latest supported Python, macos-latest)
+          macos: py3
+          python-version: ${{ needs.supported-pythons.outputs.latest }}
+      artifact-path: |
+        results.xml
```
braingram, line 83: Which run produces this file?

.github/workflows/tests.yml
```diff
@@ -1,46 +1,77 @@
...
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions: {}
+
+env:
+  PYTHON_VERSIONS: 3.9 3.10 3.11 3.12 3.13 3.14
```
braingram, line 27: I'm a bit confused by this and the use of the open astronomy workflow. Doesn't that already support globbing over the python version?
When 3.15 is stable won't we need to edit this workflow to add 3.15? If we're already editing this file why not just list the environments?

.github/workflows/tests.yml
```diff
@@ -1,46 +1,77 @@
...
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@e5af21e4ae37af089a73bb032e10b78b0e746174 # v3.0.1
     with:
-      setenv: |
-        CRDS_PATH: /tmp/data/crds_cache
-        CRDS_CLIENT_RETRY_COUNT: 3
-        CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
       envs: |
-        - linux: py313-jwst
-        - linux: py313-romancal
+        - name: run tests (oldest supported Python, ubuntu-latest)
```
braingram, line 70: Where are:
- oldestdeps
- downstreamdeps
- nolegacypath
- xdist factor

.pre-commit-config.yaml
```diff
@@ -17,10 +18,58 @@ repos:
       - id: check-toml
       - id: check-merge-conflict
       - id: check-symlinks
-      # - id: debug-statements
+      - id: debug-statements
```
braingram: This will require code changes. Would you re-comment it out?

.pre-commit-config.yaml
```diff
@@ -30,48 +79,28 @@ repos:
...
+      - id: numpydoc-validation
+        exclude: |
+          (?x)^(
+            jwst/tests/.* |
+            docs/.* |
+            jwst/regtest/test_.* |
+            .*/tests/test_.*
+          )$
+  # C style
+  - repo: https://github.com/pre-commit/mirrors-clang-format
```
braingram: This repo contains no c code. Would you remove this hook?

.pre-commit-config.yaml
```diff
@@ -30,48 +79,28 @@ repos:
...
-  - repo: https://github.com/twisted/towncrier
-    rev: 25.8.0
+      - id: numpydoc-validation
+        exclude: |
+          (?x)^(
+            jwst/tests/.* |
+            docs/.* |
+            jwst/regtest/test_.* |
+            .*/tests/test_.*
+          )$
```
braingram: These paths don't match the repository.

.pre-commit-config.yaml
```diff
@@ -30,48 +79,28 @@ repos:
...
-  # # type checking
-  # - repo: https://github.com/pre-commit/mirrors-mypy
-  #   rev: v1.17.0
-  #   hooks:
-  #     - id: mypy
-  # spell-checking
-  - repo: https://github.com/codespell-project/codespell
-    rev: v2.4.2
+  # validate numpy-style docstrings
+  - repo: https://github.com/numpy/numpydoc
```
braingram: Would you comment out this hook? If we want to enable it we'll need code/doc changes.

.github/release.yml
braingram, line 1: What is this file for?

changes/README.md
```diff
@@ -0,0 +1,22 @@
...
+and insert them as a full changelog entry at the top of `CHANGES.rst` for the released version.
+
+News fragment filenames consist of the pull request number and the changelog category (see below).
+A single change can have more than one news fragment, if it spans multiple categories.
+
+## Change Log Categories
+
+Make a news fragment for every relevant category affected by your change.
+
+- `<PR#>.breaking.rst`: Add this fragment if your change **breaks public API**, describing what the user needs to change
```
braingram: Why not refer to the configuration (pyproject.toml or towncrier.toml) instead of duplicating the categories here?

docs/README.md
```diff
@@ -0,0 +1,18 @@
+# Writing and maintaining documentation
+
+Documentation for `stpipe` is written in [Sphinx reStructuredText (`.rst`)](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html)
+in this `docs/` directory, and is hosted online athttps://stpipe.readthedocs.io### Building documentation locally
```
braingram:
```suggestion
in this `docs/` directory, and is hosted online at https://stpipe.readthedocs.io
### Building documentation locally
```
Formatting does not look correct here.

.gitignore
```diff
@@ -139,7 +102,15 @@ dmypy.json
...
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+.idea/
+
+# ruff
+.ruff_cache/
 
-docs/source/api
```
braingram, line 145: Is this still in the .gitignore?

.pre-commit-config.yaml
braingram, line 1: There are a number of new checks in this configuration. I can see the benefit in trying to standardize across a set of checks but doing that as part of the template application isn't going to make these PRs easy to review and will delay their merge.

How about we don't enable new checks as part of this PR and instead comment out or remove the new checks and open issues/separate PRs to enable those?

.readthedocs.yaml
```diff
@@ -12,31 +11,22 @@ build:
     - graphviz
   jobs:
     post_checkout:
-      - git fetch --unshallow || true
```
braingram, line 15: Docs are failing and I suspect these changes are related. Some of the checks/conditionals here were added to address issues that only pop up at release time/after a release. It's my understanding that the steps/checks are similar across repositories (or should be) so why are they mostly being removed?

For example, IIRC:
```
git update-index --assume-unchanged docs/source/conf.py
```
was added to address incorrect documentation versions for releases. Won't removing this reintroduce the issue?

.ruff.toml
```diff
@@ -1,31 +1,68 @@
...
 ]
-line-length = 88
 extend-exclude = [
   "docs",
   "scripts/strun",
 ]
 
 [lint]
 select = [
+  # "YTT",  # flake8-2020
```
braingram: Would you remove these comments? They clutter the file and there are no plans to enable all these rules.

.ruff.toml
```diff
@@ -1,31 +1,68 @@
...
   "UP",  # pyupgrade
+  # "FURB",  # refurb
+  # "RUF",  # Ruff-specific rules
+  # "TRY",  # tryceratops
 ]
 extend-select = [
   "NPY",  # NumPy-specific checks (recommendations from NumPy)
 ]
+ignore = [
+  "D100",  # Missing docstring in public module
```
braingram: Why are we ignoring rules for sets we don't enable?

pyproject.toml
```diff
@@ -84,25 +84,3 @@ force-exclude = "^/(\n  (\n      \\.eggs\n    | \\.git\n    | \\.pytest_cache\n
 [tool.codespell]
 skip = "*.pdf,*.fits,*.asdf,.tox,build,./tags,.git,docs/_build"
 
-[tool.towncrier]
```
braingram, line 87: Why don't we keep this configuration here? Isn't that the point of the `tool` sections in this file?

README.md
```diff
@@ -1,9 +1,16 @@
+<a href="https://stsci.edu">
+  <img src="docs/_static/stsci_pri_combo_mark_horizonal_white_bkgd.png" alt="Space Telescope Science Institute" width="83%" style="margin-left: auto;"/>
```
braingram: Won't this incorrectly render on pypi? Rather than duplicating these images in every repository can we link to:
- an existing public location?
- the package template?

README.md
```diff
@@ -1,9 +1,16 @@
...
+</a>
+
 # stpipe
 
-[![CI](https://github.com/spacetelescope/stpipe/actions/workflows/tests.yml/badge.svg)](https://github.com/spacetelescope/stpipe/actions/workflows/tests.yml)
-[![codecov](https://codecov.io/gh/spacetelescope/stpipe/branch/main/graph/badge.svg?token=Mm9I0X1o4X)](https://codecov.io/gh/spacetelescope/stpipe)
+[![build](https://github.com/spacetelescope/stpipe/actions/workflows/build.yml/badge.svg)](https://github.com/spacetelescope/stpipe/actions/workflows/build.yml)
+[![tests](https://github.com/spacetelescope/stpipe/actions/workflows/tests.yml/badge.svg)](https://github.com/spacetelescope/stpipe/actions/workflows/tests.yml)
+[![ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
```
braingram:
```suggestion
```
This feels like free advertising with no benefit to users.

.github/workflows/label.yml
zacharyburnett: if we label PRs then we can also auto-categorize the release notes with `.github/release.yml`

.github/workflows/tests.yml
zacharyburnett: oh that's a good point. We can just manually set oldest and latest Python 😅 I overcomplicated it

.github/workflows/label.yml
braingram: Why don't we use the changelog as the release notes?

.github/workflows/build.yml
zacharyburnett: good catch, this needs to stay in

.github/workflows/label.yml
zacharyburnett: We can, but you have to convert from restructured text to Markdown every time (every link). That's what I've been doing manually for recent release notes, but I think that might be burdensome to expect anyone else to do. Our other alternative is we just use the auto release notes of uncategorized PRs in a big list

.github/workflows/tests.yml
zacharyburnett: I figured if it's running on `main` already it will run on tags, but now that I think about it not all tags are on `main`. I'll restore this too

.github/workflows/tests.yml
zacharyburnett: I wanted to discuss this as well; do we need scheduled runs for unit test CI? I don't know if anyone is actually looking at these. I can restore this and we can make a PR to remove them to ask the devs

braingram:
@zacharyburnett this has docs and pre-commit failures?

Is it possible to split this into a smaller set of changes so we can start using the template for the AI policy at least?

braingram (changes requested):
From offline communication these changes were broken up:
- https://github.com/spacetelescope/stpipe/pull/371 linked this repo to the template and added an AI_POLICY.md
- follow-up PR(s) will apply further changes

This was done to hasten the inclusion of the policy and introduce the syncing (so we can see how it behaves). Syncing was tested here https://github.com/spacetelescope/stpipe/actions/runs/34371844215 resulting in https://github.com/spacetelescope/stpipe/pull/371.

Some issues to sort out for further updates:
- The template does not have an auto-update github actions workflow versions since these live in jinja templates. We should sort this out before rolling out syncing actions and disabling dependabot.
- It is unclear if we want copies of CONTRIBUTING.md TESTING.md in every repository or have it in a centralized location. It would be helpful to see how the files in template match the existing versions of these files in the pipelines.
- The PR template in this PR removes the "Resolves RCAL ..." line. To match the current PR template for this repo both JP and RCAL projects should be listed.

For this PR in particular:
- main has a LICENSE file, this PR adds LICENSE.txt but does not remove LICENSE
- the codecov badge is removed from README.md. It should be re-added.

PR #311, opened 2026-04-09
[DOCS] use `pip` and `apt` to install packages for RTD builds

zacharyburnett:
## Tasks

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,26 +1,25 @@
-# .readthedocs.yaml
-# Read the Docs configuration file
-# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+# yaml-language-server: $schema=https://raw.githubusercontent.com/readthedocs/readthedocs.org/refs/heads/main/readthedocs/rtd_tests/fixtures/spec/v2/schema.json
+# https://docs.readthedocs.io/en/stable/config-file/v2.html
 
 # Required
 version: 2
 
 build:
-  os: ubuntu-24.04
+  os: ubuntu-lts-latest
   tools:
-    python: mambaforge-latest
+    python: latest
+  apt_packages:
+    - graphviz
   jobs:
     post_checkout:
       - git fetch --unshallow || true
       - git config remote.origin.fetch '+refs/heads/*:refs/remotes/origin/*' || true
       - git fetch --all --tags || true
     pre_install:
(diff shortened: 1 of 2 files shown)


.readthedocs.yaml
```diff
@@ -1,27 +1,26 @@
...
+  apt_packages:
+    - graphviz
   jobs:
     post_checkout:
       - git fetch --unshallow || true
       - git config remote.origin.fetch '+refs/heads/*:refs/remotes/origin/*' || true
       - git fetch --all --tags || true
     pre_install:
-      - git update-index --assume-unchanged docs/rtd_environment.yaml docs/source/conf.py
+      - git update-index --assume-unchanged docs/source/conf.py
```
braingram, line 19: Is the goal to remove?
https://github.com/spacetelescope/stpipe/blob/main/docs/rtd_environment.yaml
If so would you remove it in this PR?

.readthedocs.yaml
```diff
@@ -1,27 +1,26 @@
-# .readthedocs.yaml
-# Read the Docs configuration file
-# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+# yaml-language-server: $schema=https://raw.githubusercontent.com/readthedocs/readthedocs.org/refs/heads/main/readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```
braingram, line 1: What does this comment do?

.readthedocs.yaml
zacharyburnett: oh yes, thanks!

.readthedocs.yaml
zacharyburnett: this tells the currently-running YAML language server to explicitly use this schema for validation in the editor while editing, to hopefully catch simple errors before uploading it to a build on RTD

it works for Helix and should work for vscode as well, and also nvim if you have a language server set up

braingram: Thanks. Is there any way to add this to pre-commit or in some other way use it in the CI (or use pre-commit to achieve something similar)? To expand on why I ask, I don't have anything in my local environment that will use this comment. Since the CI also won't check it I could easily delete or modify this line and see no errors or changes but you might start seeing issues.

zacharyburnett: oh very good point, I'll look into schema validators hooks we can add

braingram (approved):
Thanks!

PR #315, opened 2026-05-11
remove unused cmdline.step_script

braingram:
Closes https://github.com/spacetelescope/stpipe/issues/178

Remove unused cmdline.step_script

Unused in jwst: https://github.com/search?q=repo%3Aspacetelescope%2Fjwst%20step_script&type=code
and unused in romancal: https://github.com/search?q=repo%3Aspacetelescope%2Fromancal+step_script&type=code

jwst regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/25672081673
romancal regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/25672114547

This removes the only use of the `cls` argument to `step_from_cmdline`:
https://github.com/spacetelescope/stpipe/blob/14c962784fcbcc58ad142918227ac530b313391b/src/stpipe/cmdline.py#L431
(FWIW `step_from_cmdline` is unused outside of stpipe). If we remove `cls` it would remove all non-test uses of the `cls` argument to `just_the_step_from_cmdline` (FWIW `just_the_step_from_cmdline` is unused outside of tests). I think we should simplify these functions by:
- removing the unused arguments
- replacing test usage of `just_the_step_from_cmdline` with `step_from_cmdline`
- making `just_the_step_from_cmdline` private (and perhaps removing it depending on how we want to address capturing the pre-pipeline jobs)

I didn't include the above changes in this PR but can do so if it's preferred to batch these changes into 1 "refactoring" PR.

## Tasks
- [ ] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [ ] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see [changelog readme](https://github.com/spacetelescope/stpipe/blob/main/changes/README.rst) for instructions)
  - [ ] run regression tests with this branch installed (`stpipe@git+https://github.com/<fork>/stpipe.git@<branch>`)
    - [ ] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml)
    - [ ] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml)

PR #309, opened 2026-04-07
JP-4129: Capture pre-pipeline CRDS checks in ASDF log (try 2)

pllim:
Resolves [JP-4129](https://jira.stsci.edu/browse/JP-4129)

* This PR addresses https://github.com/spacetelescope/jwst/issues/9864
* Close https://github.com/spacetelescope/stpipe/pull/304

Blocked by:

* https://github.com/spacetelescope/stpipe/pull/312
* https://github.com/spacetelescope/jwst/pull/10523
* #315 
* #317 

TODO:

- [ ] Fix the problem in `strun`
- [ ] Write tests

## Tasks
- [ ] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [x] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [x] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see [changelog readme](https://github.com/spacetelescope/stpipe/blob/main/changes/README.rst) for instructions)
  - [x] run regression tests with this branch installed (`stpipe@git+https://github.com/pllim/stpipe.git@log-all-the-things-porque-no-los-dos`)
    - [x] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml)
      - https://github.com/spacetelescope/RegressionTests/actions/runs/24103349581 (passed with spacetelescope/jwst#10437)
      - https://github.com/spacetelescope/RegressionTests/actions/runs/25520758464
    - [x] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml)
      - https://github.com/spacetelescope/RegressionTests/actions/runs/24103582716 (passed)
      - https://github.com/spacetelescope/RegressionTests/actions/runs/25520810950

PR #317, opened 2026-05-18
capture pre-pipeline/step logs and cleanup cmdline

braingram:
https://github.com/spacetelescope/stpipe/pull/315 removed the unused `step_script` method from `cmdline`. It's removal allows for some cleanup in cmdline including:
- removal of the now un-used `cls` method from `just_the_step_from_cmdline` (more on this below) which allows removal of `cls` from `_build_parent_arg_parser`
- `just_the_step_from_cmdline` is only (directly) used in stpipe unit tests, there is no romancal or jwst usage, at the moment this PR removes it (and replaces the usage in the unit tests with the used `step_from_cmdline`)
- the above allows some cleanup of the argparsers in cmdline including
  - removal of `apply_log_config` arguments (also removed in all cmdline methods)
  - split parsing of logs/class (to allow easier integration of logging to capture pre-pipeline logs)
- the above changes (mainly the easier logging integration) allowed some cleanup of the logging
  - removal of `record_logs` (this is now handled by `LogConfig`)
  - require `log_names` for `apply` and `undo`
  - removal of `close_handlers` argument to `undo`

These changes also allow attaching the log recording handler earlier in the step/pipeline execution process allowing pre-pipeline logs to be captured for strun and call usage.

On main running `Step.call(configure_log=False)` fails to capture warnings in the recorded logs (following https://github.com/spacetelescope/stpipe/pull/271). This PR retains that behavior (see https://github.com/spacetelescope/stpipe/pull/317/commits/28aeeb11ce436aac438112677217339804286ccc). If we want to fix that issue, we'll need to update one jwst test:
https://github.com/spacetelescope/jwst/blob/4728a31f536c3339b33f2ef694d41d9616644d37/jwst/emicorr/tests/test_emicorr.py#L353
since the captured deprecation warning won't get captured by pytest.warns (due to a poor interaction between pytest and logging). If we revert the above commit and remove the `pytest.warns` (which will need to be replaced by a filterwarnings for the message to avoid the uncaught warnings turning into an error) the warning will correctly appear twice in the log (since jwst emits identical warning and a log.warning messages). On main (and with this PR with the above commit) the warning message appears only once in the cal_logs due to only the log.warning message being captured.

regtests
romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/26038917156
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/26038956800

## Tasks

### DIFF
--- a/src/stpipe/cmdline.py
+++ b/src/stpipe/cmdline.py
@@ -289,7 +289,7 @@ def _determine_log_configuration(known):
 
     Returns
     -------
-    log_cfg : logging configuration object
+    log_cfg : LogConfig
         The loaded logging configuration ready for use.
     """
     if known.logcfg is not None:
--- a/src/stpipe/cmdline.py
+++ b/src/stpipe/cmdline.py
@@ -230,7 +230,7 @@ def _override_config_from_args(config, args):
             set_value(config, key, val)
 
 
-def just_the_step_from_cmdline(args, cls=None, apply_log_cfg=False):
+def just_the_step_from_cmdline(args, apply_log_cfg=False):
     """
     Create a step from a configuration file and return it.  Don't run it.
 
@@ -240,10 +240,6 @@ def just_the_step_from_cmdline(args, cls=None, apply_log_cfg=False):
     ----------
     args : list of str
         Command line arguments
-
-    cls : Step class, optional
-        The step class to use.  If none is provided, the step is inferred
-        from the input arguments.
 
     apply_log_cfg : bool
         If True, apply the logging configuration. If False,
@@ -261,7 +257,7 @@ def just_the_step_from_cmdline(args, cls=None, apply_log_cfg=False):
         instance.
 
     step_class: Step class
-        As defined by `cls` parameter or .cfg file.
+        As defined by .cfg file.
 
     positional: list of strings
         Positional parameters after arg parsing
@@ -269,21 +265,13 @@ def just_the_step_from_cmdline(args, cls=None, apply_log_cfg=False):
     debug_on_exception : bool
         If True, the Python debugger will be invoked when an exception occurs.
     """
-    parser1 = _build_parent_arg_parser(cls, apply_log_cfg=apply_log_cfg)
+    parser1 = _build_parent_arg_parser(None, apply_log_cfg=apply_log_cfg)
     known, _ = parser1.parse_known_args(args)
 
     try:
-        if cls is None:
-            step_class, config, name, config_file = _get_config_and_class(
-                known.cfg_file_or_class[0]
-            )
-        else:
-            config_file = known.config_file
-            config = config_parser.load_config_file(config_file)
-            step_class, name = Step._parse_class_and_name(
-                config, config_file=config_file
-            )
-            step_class = cls
+        step_class, config, name, config_file = _get_config_and_class(
+            known.cfg_file_or_class[0]
+        )
 
         if apply_log_cfg:
             # Retrieve a log config file if specified
@@ -352,10 +340,7 @@ def just_the_step_from_cmdline(args, cls=None, apply_log_cfg=False):
 
     args = parser2.parse_args(args)
 
-    if cls is None:
-        del args.cfg_file_or_class
-    else:
-        del args.config_file
+    del args.cfg_file_or_class
     del args.debug
     del args.save_parameters
     del args.disable_crds_steppars
--- a/src/stpipe/cmdline.py
+++ b/src/stpipe/cmdline.py
@@ -222,97 +222,66 @@ def _override_config_from_args(config, args):
             set_value(config, key, val)
 
 
-def _just_the_step_from_cmdline(args):
-    """
-    Create a step from a configuration file and return it.  Don't run it.
+def _print_parser_error(parser, error):
+    _print_important_message("ERROR PARSING CONFIGURATION:", str(error))
+    parser.print_help()
 
-    DOES NOT RUN THE STEP.
 
-    Parameters
-    ----------
-    args : list of str
-        Command line arguments
+def _config_and_class_from_cmdline(args):
+    parser = _build_parent_arg_parser()
+    known, _ = parser.parse_known_args(args)
+    try:
+        return (*_get_config_and_class(known.cfg_file_or_class[0]), parser, known)
+    except Exception as e:
+        _print_parser_error(parser, e)
+        raise e
 
-    Returns
-    -------
-    step : Step instance
-        If the config file has a `class` parameter, or the commandline
-        specifies a class, the return value will be as instance of
-        that class.
 
-        Any parameters found in the config file or on the commandline
-        will be set as member variables on the returned `Step`
-        instance.
+def _determine_log_configuration(known):
+    if known.logcfg is not None:
+        msg = (
+            "The logcfg configuration file is deprecated. "
+            "Please use the log_* command line "
+            "arguments to configure logging."
+        )
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
 
-    step_class: Step class
-        As defined by .cfg file.
+    # if verbose is enabled
+    # or log_level log_file or log_stream are set (not None)
+    # then don't try to use the log configuration file (or any passed in log_cfg)
+    if known.verbose is True or any(
+        getattr(known, attr) is not None
+        for attr in ("log_level", "log_file", "log_stream")
+    ):
+        cfgfile = None
+    elif known.logcfg:
+        if not os.path.exists(known.logcfg):
+            raise OSError(f"Logging config {known.logcfg!r} not found")
+        cfgfile = known.logcfg
+    else:
+        cfgfile = log._find_logging_config_file()
 
-    positional: list of strings
-        Positional parameters after arg parsing
-
-    debug_on_exception : bool
-        If True, the Python debugger will be invoked when an exception occurs.
-    """
-    parser1 = _build_parent_arg_parser()
-    known, _ = parser1.parse_known_args(args)
+    # determine level
+    if known.verbose:
+        log_level = "DEBUG"
+    elif known.log_level is not None:
+        log_level = str(known.log_level).upper()
+    else:
+        log_level = None
 
     try:
-        step_class, config, name, config_file = _get_config_and_class(
-            known.cfg_file_or_class[0]
+        log_cfg = log.load_configuration(
+            config_file=cfgfile,
+            log_level=log_level,
+            log_file=known.log_file,
+            log_stream=known.log_stream,
         )
-
-        # Retrieve a log config file if specified
-        use_log_cfg = True
-        log_args = [
-            known.log_level,
-            known.log_file,
-            known.log_stream,
-        ]
-        if known.verbose is True or any([arg is not None for arg in log_args]):
-            use_log_cfg = False
-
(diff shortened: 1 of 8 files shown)


src/stpipe/cmdline.py
```diff
@@ -65,25 +65,18 @@ def _get_config_and_class(identifier):
...
-    if cls is None:
-        parser1.add_argument(
-            "cfg_file_or_class",
-            type=str,
-            nargs=1,
-            help="The configuration file or Python class to run",
-        )
-    else:
-        parser1.add_argument(
-            "--config-file",
```
pllim, line 83: So `--config-file` is no longer an option. Is this considered major API change?

pllim (approved):
I would support any simplification for the CLI stuff. Thanks! Approval FWIW, not that it matters in this repo.

src/stpipe/cmdline.py
braingram: Thanks for pointing this out! `--config-file` handling is a little special.

For strun, `--config-file` isn't an option on main (it's expected that the config file be provided as a positional argument).

For Step.from_cmdline `--config-file` also isn't supported (same reason).

It was possible to trigger `--config-file` as an option with `step_script` (which provided `cls` to `step_from_cmdline`, which then provided it to `just_the_step_from_cmdline` which then provided it to this method).
`just_the_step_from_cmdline` and `step_from_cmdline` aren't mentioned in the docs (nor is anything in the `cmdline` submodule). https://github.com/spacetelescope/stpipe/pull/318 is work towards making the public vs private parts of the API more clear and renames this module to `_cmdline`.

pllim: I wonder if this should be mentioned in the change log but then again it is not supposed to be public... 🤷‍♀️ 

Thanks for the clarification!

braingram: Sounds good to me. How does 1e99658 look?

pllim: LGTM. Thanks!

src/stpipe/step.py
```diff
@@ -490,9 +490,23 @@ def run(self, *args):
...
-        with log.record_logs(
-            log_names=self.get_stpipe_loggers(), formatter=self._log_records_formatter
-        ) as log_records:
+        # if run is called directly attach handlers to record logs
+        if log.LogConfig.applied is None:
+            ctx = log.LogConfig(
+                [],
+                level=logging.NOTSET,
+                recording_formatter=self._log_records_formatter,
+            ).context(
```
melanieclarke, line 499: I'm trying to understand the implications of the changes here.  Mostly, I really like moving the recording handling to the log configuration, but I'm wondering about this change specifically.

I think using the context manager here will call LogConfig.apply, which sets the class attribute `applied`, meaning that any concurrent `run` calls will pick up the log config applied here, even if there wasn't one applied previously.  This makes the `log_records` not thread safe, where before, I think it should have been possible to configure the log separately, run multiple steps at once via `run` (or `call` with `configure_log=False`), and still record logs in data products correctly.

I don't know that our steps are otherwise thread-safe, so maybe this is a pointless discussion, but I did try to design the new log handling so that it was at least possible for expert users to set up advanced logging use cases.  Is it possible to avoid using the class attribute from inside `run`?  Or are we deciding we just don't support threading?

braingram: Thanks. I looks like we already have cross-thread issues with the logging on main. Do we have any tickets about using stpipe/jwst in a multithreaded way? Running 2 instances of jwst dqinitstep in 2 threads results in identical log messages in the 2 resulting files. Forcing custom logs per run (a log message "index=0" for thread 0 and "index=1" for thread 1) both output files contain both log messages.

While working on this PR I considered using thread local storage for the configuration but I was worried it was making this PR too large (and addressing too much). One issue I'm not sure how we'll get around is that all threads share loggers (calling `logging.getLogger("stdatamodels")` will return the same object in all threads). If we want to separate logs from different threads and allow each their own configuration I think we may need to rethink how we're configuring logging.

Here's the toy example, you'll need to update the paths for your system:
<details>

```python
import concurrent.futures
from jwst.dq_init import DQInitStep

import logging

fns = [
    "~/test_bigdata/jwst-pipeline/dev/nircam/image/jw01538046001_03105_00001_nrcalong_uncal.fits",
    "~/test_bigdata/jwst-pipeline/dev/nircam/image/jw00634_nrcblong_no_mtt_uncal.fits",
]

class MyStep(DQInitStep):
    def __init__(self, index, *args, **kwargs):
        self._index = index
        super().__init__(*args, **kwargs)

    def process(self, *args, **kwargs):
        logging.getLogger("stdatamodels").warning(f"{self._index=}")
        return super().process(*args, **kwargs)


steps = [MyStep(i) for i, fn in enumerate(fns)]


def run_step(step, fn):
    print(id(logging.getLogger("stdatamodels")))
    return step.run(fn)


with concurrent.futures.ThreadPoolExecutor(max_workers=len(fns)) as executor:
    results = executor.map(run_step, steps, fns)


for model in results:
    print(model.cal_logs.dq_init.instance)
```
</details>

melanieclarke: Got it, thanks for looking into that.  I don't know of any tickets asking for thread-safe logging, so it seems reasonable to me to defer dealing with it until there's demand for it.

Should we add a note to the docs somewhere that the stpipe logging infrastructure is not thread-safe?

braingram: Thanks. Should we open a ticket for it?

I would say it's "thread safe" in so far as it doesn't crash so if we do add a note I suggest we clarify that the logs will contain messages from all threads (unless we know of other issues).

melanieclarke: Sounds good on both counts.  I don't know of any crashing issues, it just seems like a good idea to document that it won't work as expected.

melanieclarke:
Testing locally, the command line case looks great. I see the start up messages from CRDS, on screen and in the datamodel cal_logs.  If I set log-stream=null, I see no messages in the terminal, as expected, including the CRDS messages that still appear on main.  The cal_logs are still populated correctly in this case.

From a Python session, the default log config case also looks good, but I'm seeing something slightly weird with the CRDS logger with configure_log=False.  Here's a minimal example to reproduce, with a file from the regression tests:
```
from jwst.assign_wcs import AssignWcsStep

model = AssignWcsStep.call('jw01538046001_03105_00002_nrcalong_rate.fits', configure_log=False)
print("\n".join(model.cal_logs.assign_wcs))

model = AssignWcsStep.call('jw01538046001_03105_00002_nrcalong_rate.fits', configure_log=False)
print("\n".join(model.cal_logs.assign_wcs))
```

The first call prints a CRDS message and stores it in the cal_logs.  The second one prints no messages and the cal_logs are empty.  This isn't a show-stopper, and it looks like it behaves similarly on main, so not urgent to fix here. I just wanted to mention it because it came up in testing here.

braingram:
> From a Python session, the default log config case also looks good, but I'm seeing something slightly weird with the CRDS logger with configure_log=False. Here's a minimal example to reproduce, with a file from the regression tests:
> 
> ```
> from jwst.assign_wcs import AssignWcsStep
> 
> model = AssignWcsStep.call('jw01538046001_03105_00002_nrcalong_rate.fits', configure_log=False)
> print("\n".join(model.cal_logs.assign_wcs))
> 
> model = AssignWcsStep.call('jw01538046001_03105_00002_nrcalong_rate.fits', configure_log=False)
> print("\n".join(model.cal_logs.assign_wcs))
> ```
> 
> The first call prints a CRDS message and stores it in the cal_logs. The second one prints no messages and the cal_logs are empty. This isn't a show-stopper, and it looks like it behaves similarly on main, so not urgent to fix here. I just wanted to mention it because it came up in testing here.

Thanks for testing this an sharing the example. When I run it locally I don't see any log messages for main or this PR which is even more puzzling. I'm running:
- python 3.12
- jwst ebdcd33b1c1dc33f8faadca7ad86fc240c9ba121 (latest commit on main)
- crds 13.2.1
- with environment variables:
  - CRDS_CONTEXT=jwst-latest
  - CRDS_PATH=/Users/bgraham/crds_cache/jwst_ops
  - CRDS_SERVER_URL=https://jwst-crds.stsci.edu

How does this compare to the environment you used for testing? Also what are the CRDS log messages you saw?

melanieclarke:
Try unsetting CRDS_CONTEXT.  The message I see is:
2026-05-22 09:25:34,322 - CRDS - INFO -  DEV calibration SW identified. Defaulting to Edit Context.

braingram:
> Defaulting to Edit Context

Thanks! That message should only appear once since it's triggered by get_processing_mode in crds (which is cached).

melanieclarke:
Nice, thanks for tracking that down!  I think all behavior is as expected, then.

melanieclarke (commented):
These changes look good to me, and testing looks like it is working as expected, including capturing the start up messages from stpipe and CRDS.  This should close JP-4129, and can replace #309.

Two small documentation requests:
- I think we should add a note to the logging docs for stpipe that the logging infrastructure won't work correctly if steps are run concurrently in the same Python process.
- Can you please add some docstrings to the new internal functions in `cmdline` (`_print_parser_error`, `_config_and_class_from_cmdline`, etc)?  They're not required for private functions, but might help the next developer trying to understand what's going on in this module.

braingram:
I asked copilot to generate some docstrings and with some minor cleanup they looked reasonable to me. How do these look?
https://github.com/spacetelescope/stpipe/pull/317/commits/faed4f8000c8cd2296ade36322a5aabe436e5c55

For updating the docs to mention multi-threaded behavior can we split this into a separate PR/issue since this PR does not change the behavior (as far as we've tested it)? It seems worthwhile to first do more testing on multi-threaded uses or perhaps to make the note more general saying that stpipe is not designed for or tested with multi-threaded uses.

src/stpipe/cmdline.py
```diff
@@ -230,114 +222,151 @@ def set_value(subconf, key, val):
...
+        - verbose: Enable all logging messages
+        - log_level: Specific log level to set
+        - log_file: Path to log file
+        - log_stream: Output stream for logs
 
-    debug_on_exception : bool
-        If True, the Python debugger will be invoked when an exception occurs.
+    Returns
+    -------
+    log_cfg : logging configuration object
```
melanieclarke: This is a nitpick for docs on a private function so feel free to ignore, but this should document the actual type of the returned object.  

```suggestion
    log_cfg : LogConfig
```

src/stpipe/cmdline.py
```diff
@@ -230,114 +222,151 @@ def set_value(subconf, key, val):
...
+    parser : argparse.ArgumentParser
+        The parent argument parser.
+    known : argparse.Namespace
+        Parsed parent-level command line arguments.
+    args : list of str
+        Remaining unparsed command line arguments.
 
+    Returns
+    -------
+    step : Step instance
```
melanieclarke: Here too, it should specify the type.
```suggestion
    step : Step
```

melanieclarke (commented):
Okay by me to defer the docs for the threading issues.  Can you please file the ticket though, so we don't lose it?

melanieclarke (approved):

src/stpipe/step.py
braingram: I opened https://github.com/spacetelescope/jwst/issues/10578

Feel free to update it if it's missing details.

PR #318, opened 2026-05-20
Underscore private modules

braingram:
Rename a number of undocumented submodules adding an underscore prefix.

Some items to untangle by updating jwst unit tests
- config_parser (~1 mention in jwst docs~ EDIT: docs updated, can deprecate in 1.x once jwst switches imports of ValidationError to stpipe.exceptions)
- entry_points (~1 use in romancal tests~ EDIT: test usage removed can deprecate in 1.x)
- utilities (1 use in romancal tests)

~format_template is used by association code in jwst and romancal (although this is likely functionally unused but imported code in romancal: https://github.com/spacetelescope/romancal/issues/2326).~ EDIT: format_template is now moved to jwst, the romancal usage can be removed

romancal regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/26175812659 all pass
jwst regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/26175797950 show 2 unrelated failures (same seen in unrelated PRs)

## Tasks


pllim:
I am neutral on putting underscore in an entire module. I thought there was a larger discussion somewhere on documenting what is public vs private API but I cannot find it. Do you remember if that actually is an issue somewhere or did I confuse it with Astropy stuff, @emolter or @zacharyburnett ?

I could only find https://github.com/spacetelescope/jwst/pull/10096 that was very specific to `jwst`.

emolter:
@pllim I don't recall that discussion

src/stpipe/config_parser.py
```diff
@@ -20,26 +20,23 @@
...
 from .datamodel import AbstractDataModel
+from .exceptions import ValidationError
 from .utilities import _not_set
 
 # Configure logger
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
 
-class ValidationError(Exception):
```
braingram, line 32: Moved to exceptions.py (with the other exceptions). FWIW jwst imports this exception so after this PR is merged we can update jwst to import this from exceptions.

braingram:
> I am neutral on putting underscore in an entire module. I thought there was a larger discussion somewhere on documenting what is public vs private API but I cannot find it. Do you remember if that actually is an issue somewhere or did I confuse it with Astropy stuff, @emolter or @zacharyburnett ?
> 
> I could only find [spacetelescope/jwst#10096](https://github.com/spacetelescope/jwst/pull/10096) that was very specific to `jwst`.

I'd like this package to follow pep8. Here's a relevant section: https://peps.python.org/pep-0008/#public-and-internal-interfaces (with the relevant text here with minor formatting to make github markdown behave)

> Even with ``__all__`` set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

src/stpipe/config_parser.py
emolter: and this doesn't break jwst temporarily because `ValidationError` is still technically available from here, even after removing it, because it's imported here?

tests/test_logger.py
```diff
@@ -307,7 +306,7 @@ def test_from_cmdline_no_root_logger_changes_level_arg(
 
 def test_step_from_cmdline_no_root_logger_changes(log_cfg_path, root_logger_unchanged):
     with pytest.warns(DeprecationWarning, match="logcfg configuration file"):
-        stpipe.cmdline.step_from_cmdline(
```
emolter, line 310: What's the difference between `stpipe.cmdline.step_from_cmdline` and `Step.from_cmdline`, and why did you change it to the latter in most tests below, but not in this one?

emolter (approved):
I'd like this to get some additional reviews, so please don't merge yet, but IMO this can only be an improvement.

src/stpipe/config_parser.py
braingram: Exactly. Moving the defintion doesn't prevent jwst from importing `ValidationError` from here because this file import `ValidationError`. Once this PR is merged we should update jwst to import `ValidationError` from `stpipe.exceptions`.

tests/test_logger.py
braingram: `Step.from_cmdline` is essentially an alias of `stpipe.cmdline.step_from_cmdline`:
https://github.com/spacetelescope/stpipe/blob/416bcf9492c4fbe452a127767a9c908074b6d587/src/stpipe/step.py#L185-L187
`Step.from_cmdline` is used extensively outside of stpipe whereas `stpipe.cmdline.step_from_cmdline` is only used internally in stpipe.
There is another test that already checks `Step.from_cmdline`:
https://github.com/spacetelescope/stpipe/blob/416bcf9492c4fbe452a127767a9c908074b6d587/tests/test_logger.py#L292-L296

We could remove this test but since both tests exist I figured updating this one to test the private API made sense. I can't say I understand the design decisions here in stpipe but I'm assuming that `Step.from_cmdline` exists to allow pipelines to override the behavior (although we've never had a need to do so). Even in that case it seems unlikely that we'd want a user to call `stpipe.cmdline.step_from_cmdline` and instead we'd want them to call `Step.from_cmdline` so it seems reasonable to me to delete these tests of the private API. I'm also ok keeping them until we remove the deprecations.

braingram (commented):
Shoot, I had these comments "pending" for several days.

tests/test_logger.py
emolter: ok sounds good to me

tapastro (approved):
Looks good to me. If we get external requests to support more public modules, we can revisit as needed.

PR #334, opened 2026-07-10
JP-2161: Allow Association object as direct input

pllim:
Resolves [JP-2161](https://jira.stsci.edu/browse/JP-2161)

This PR goes with https://github.com/spacetelescope/jwst/pull/10678 . The change log is in `jwst` PR so probably no need change log here.

## Tasks


pllim:
I will just close this PR without merge and see if there is a more elegant solution.

PR #345, opened 2026-07-24
deprecate format_template

braingram:
Deprecate stpipe.format_template (unused by jwst and romancal).

regtests:
roman: https://github.com/spacetelescope/RegressionTests/actions/runs/30090945271
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/30090956266 (failures are unrelated)

## Tasks


src/stpipe/format_template.py
```diff
@@ -1,217 +1,9 @@
...
-        else:
-            default = "{" + key + "}"
-        value = kwargs.get(key, default)
-        self._used_keys.append(key)
-        if value is None:
-            value = ""
-
-        return value
+warnings.warn(
+    "stpipe.format_template is deprecated. Use fstrings instead.",
```
melanieclarke:
```suggestion
    "stpipe.format_template is deprecated. Use f-strings instead.",
```

src/stpipe/format_template.py
```diff
@@ -1,217 +1,9 @@
-"""FormatTemplate
+import warnings
 
-Format template string allowing partial formatting.
-"""
+from ._format_template import *  # noqa: F403
```
melanieclarke, line 3: This will only import `FormatTemplate`, so it will break without warning if anyone is currently importing `CONVERSION` from stpipe.format_template.  

On main, `from stpipe.format_template import CONVERSION` works.  On this branch:
```
----> 1 from stpipe.format_template import CONVERSION

ImportError: cannot import name 'CONVERSION' from 'stpipe.format_template' (stpipe/format_template.py)
```

This is probably not terribly important to support, but in the spirit of gently deprecating the module, we could add `CONVERSION` to `__all__` for `_format_template`, and just `FormatTemplate` for the public module.

changes/345.removal.rst
```diff
@@ -0,0 +1 @@
+Deprecate ``stpipe.format_template``. Use fstrings instead.
```
melanieclarke:
```suggestion
Deprecate ``stpipe.format_template``. Use f-strings instead.
```

melanieclarke (commented):
Looks good!  A couple small suggestions.

src/stpipe/format_template.py
braingram: Thanks for testing this. I wouldn't consider `CONVERSION` as public (since it's not in `__all__`) so any user shouldn't assume stability.

melanieclarke: Okay, your call.

melanieclarke (approved):

PR #347, opened 2026-07-24
deprecate print_configspec

### LINKED ISSUE #323: Deprecate print_configspec
This is only used in a jwst test:
https://github.com/spacetelescope/jwst/blob/00a04aba0b3cd588876603fc66e42af3124c088a/jwst/stpipe/tests/test_step.py#L342

and is only 2 lines of code:
https://github.com/spacetelescope/stpipe/blob/5a830431d068198b364c77a46f1c77e41d8dc162/src/stpipe/step.py#L107-L110

Perhaps it's useful? Here's an example for jwst ResampleStep:
```python 
>>> jwst.step.ResampleStep.print_configspec()
pre_hooks = list(default=list())# List of Step classes to run before step
post_hooks = list(default=list())# List of Step classes to run after step
output_file = output_file(default=None)# File to save output to.
output_dir = string(default=None)# Directory path for output files (created if nonexistent)
output_ext = string(default='.fits')# Output file type
output_use_model = boolean(default=False)# When saving use `DataModel.meta.filename`
output_use_index = boolean(default=True)# Append index.
save_results = boolean(default=False)# Force save results
skip = boolean(default=False)# Skip this step
suffix = string(default=None)# Default suffix for output files
search_output_file = boolean(default=True)# Use outputfile define in parent step
input_dir = string(default=None)# Input directory
pixfrac = "float(min=0.0, max=1.0, default=1.0)"# Pixel shrinkage factor
kernel = "option('square','gaussian','point','turbo','lanczos2','lanczos3',default='square')"# Flux distribution kernel
fillval = string(default='NAN')# Output value for pixels with no weight or flux
weight_type = "option('ivm', 'exptime', None, default='ivm')"# Input image weighting type
output_shape = "int_list(min=2, max=2, default=None)"# [x, y] order
crpix = "float_list(min=2, max=2, default=None)"# 0-based image coordinates of the reference pixel
crval = "float_list(min=2, max=2, default=None)"# world coordinates of the reference pixel
rotation = float(default=None)# Output image Y-axis PA relative to North
pixel_scale_ratio = float(default=1.0)# Ratio of output to input pixel scale.
pixel_scale = float(default=None)# Absolute pixel scale in arcsec
output_wcs = string(default='')# Custom output WCS. Overrides other WCS parameters if provided.
single = boolean(default=False)# Resample each input to its own output grid
blendheaders = boolean(default=True)# Blend metadata from inputs into output
in_memory = boolean(default=True)# Keep images in memory
enable_ctx = boolean(default=True)# Compute and report the context array
enable_err = boolean(default=True)# Compute and report the err array
report_var = boolean(default=True)# Report the variance array
propagate_dq = boolean(default=False)# propagate DQ during resampling
pixmap_stepsize = float(default=1.0)# Interpolation step size for pixel map; interpolation is used for stepsize > 1
pixmap_order = integer(default=1)# Spline order for pixel mapping, must be 1 or 3
```
Compared to just printing `spec`:
```python
>>> print(jwst.step.ResampleStep.spec)

        pixfrac = float(min=0.0, max=1.0, default=1.0)  # Pixel shrinkage factor
        kernel = option('square','gaussian','point','turbo','lanczos2','lanczos3',default='square')  # Flux distribution kernel
        fillval = string(default='NAN')  # Output value for pixels with no weight or flux
        weight_type = option('ivm', 'exptime', None, default='ivm')  # Input image weighting type
        output_shape = int_list(min=2, max=2, default=None)  # [x, y] order
        crpix = float_list(min=2, max=2, default=None)  # 0-based image coordinates of the reference pixel
        crval = float_list(min=2, max=2, default=None)  # world coordinates of the reference pixel
        rotation = float(default=None)  # Output image Y-axis PA relative to North
        pixel_scale_ratio = float(default=1.0)  # Ratio of output to input pixel scale.
        pixel_scale = float(default=None)  # Absolute pixel scale in arcsec
        output_wcs = string(default='')  # Custom output WCS. Overrides other WCS parameters if provided.
        single = boolean(default=False)  # Resample each input to its own output grid
        blendheaders = boolean(default=True)  # Blend metadata from inputs into output
        in_memory = boolean(default=True)  # Keep images in memory
        enable_ctx = boolean(default=True)  # Compute and report the context array
        enable_err = boolean(default=True)  # Compute and report the err array
        report_var = boolean(default=True)  # Report the variance array
        propagate_dq = boolean(default=False)  # propagate DQ during resampling
        pixmap_stepsize = float(default=1.0)  # Interpolation step size for pixel map; interpolation is used for stepsize > 1
        pixmap_order = integer(default=1)  # Spline order for pixel mapping, must be 1 or 3

```
Ignoring minor formatting differences I'd say the second one is more useful since it shows the step-specific parameters.

Removing it would be a small step towards cleaning up the Step methods.

braingram:
Closes #323

Deprecate `Step.print_configspec`. Use `print(Step.spec)` instead.

Regtests all pass:
roman: https://github.com/spacetelescope/RegressionTests/actions/runs/30118417184
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/30118431964

## Tasks

### DIFF
--- a/src/stpipe/step.py
+++ b/src/stpipe/step.py
@@ -106,6 +106,11 @@ class Step:
 
     @classmethod
     def print_configspec(cls):
+        warnings.warn(
+            "print_configspec is deprecated. Use print(Step.spec) instead",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         specfile = cls.load_spec_file()
         specfile.write(sys.stdout.buffer)
 
--- a/changes/347.removal.rst
+++ b/changes/347.removal.rst
@@ -1,0 +1,1 @@
+Deprecate ``Step.print_configspec``. Use ``print(Step.spec)`` instead.
--- a/tests/test_step.py
+++ b/tests/test_step.py
@@ -1043,3 +1043,8 @@ def test_merge_pipeline_config_deprecation(tmp_path):
         cfg.write(f)
     with pytest.warns(DeprecationWarning, match="merge_pipeline_config is deprecated"):
         SimplePipe.merge_pipeline_config(cfg, str(fn))
+
+
+def test_print_configspec_deprecation():
+    with pytest.warns(DeprecationWarning, match="print_configspec is deprecated"):
+        SimpleStep.print_configspec()


src/stpipe/step.py
```diff
@@ -106,6 +106,11 @@ def load_spec_file(cls, preserve_comments=_not_set):
 
     @classmethod
     def print_configspec(cls):
+        warnings.warn(
+            "print_configspec is deprecated. Use print(Step.spec) instead",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         specfile = cls.load_spec_file()
         specfile.write(sys.stdout.buffer)
```
melanieclarke, line 116: This actually does something quite different from `print(Step.spec)`, since it loads the parent spec as well as the child spec.

For example on main:
```
>>> from jwst.stpipe import Step

>>> Step.print_configspec()
pre_hooks = list(default=list())# List of Step classes to run before step
post_hooks = list(default=list())# List of Step classes to run after step
output_file = output_file(default=None)# File to save output to.
output_dir = string(default=None)# Directory path for output files (created if nonexistent)
output_ext = string(default='.fits')# Output file type
output_use_model = boolean(default=False)# When saving use `DataModel.meta.filename`
output_use_index = boolean(default=True)# Append index.
save_results = boolean(default=False)# Force save results
skip = boolean(default=False)# Skip this step
suffix = string(default=None)# Default suffix for output files
search_output_file = boolean(default=True)# Use outputfile define in parent step
input_dir = string(default=None)# Input directory

>>> print(Step.spec)

    output_ext = string(default='.fits')  # Output file type

```

This would be closer, but it doesn't pretty-print the returned ConfigObj: `print(Step.load_spec_file())`.  

I don't know if this function is used much, but it does look to me like it's serving a purpose that isn't met by `print(Step.spec)`.  Is there a reason we need to remove it?

braingram: There are no known uses of print_configspec following https://github.com/spacetelescope/jwst/pull/10736 (removed a unit test that tested print_configspec).

The motivation for removing it is largely that it's unused code. It's also a rather awkward function (writing directly to stdout) and clutters the Step API.

Switching the recommendation to `print(Step.load_spec_file())` seems fine to me. As mentioned in the linked issue (https://github.com/spacetelescope/stpipe/issues/323) the `print(Step.spec)` output looked more helpful to me (it showed the step-specific parameters instead of mixing the step-specific and inherited ones) but it does differ.

melanieclarke: Okay, maybe suggest both then?

braingram: Suggesting both sounds like an improvement to me. I updated the message and changelog in 1026f38

melanieclarke: Works for me, thanks!

melanieclarke (approved):

PR #351, opened 2026-07-28
[SCSB-274] localize build workflow for trusted publishers

zacharyburnett:
Resolves [SCSB-274](https://jira.stsci.edu/browse/SCSB-274)

- localizes build workflow so we can set up Trusted Publishers (eventually) and not need a PyPI token
- avoids Python 3.12 `pipx` issue with upstream OpenAstronomy build workflow
- applies upcoming package template

## Tasks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,15 +1,56 @@
 name: build
 
 on:
+  pull_request:
   release:
     types: [released]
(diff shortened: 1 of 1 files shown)


braingram (approved):
Thanks! Would you update the branch protection rules after this is merged?

